PR #17, opened 2021-07-23
Delete .travis.yml

thefallentree:
now that we have github actions working, this will eliminate the travis cost


tsipa:
Hold on, please.
Right now we have this fold-unfold thing which seems like a bad ux. Let me fix it first.

PR #22, opened 2021-08-04
ci: run script to test bpftool types/options sync

qmonnet:
### Consistency checks for bpftool

This PR adds a call to test_bpftool_synctypes.py to run_selftests.sh. The objective of this script is to make sure that the different parts of bpftool that should be updated on BPF UAPI changes (type lists, help messages, documentation, bash completion) do remain in sync between them, and with the UAPI headers from the kernel. The motivation to have it in the CI is to effectively make sure that new patchsets consistently update all those parts when necessary.

Notes:
- ~The new section in run_selftests.sh is called `Testing bpftool`, but this is not entirely correct. We don't test bpftool itself (only check that different pieces of its code are in sync), whereas _other_ scripts that are part of the BPF selftests suite _do_ test bpftool, outside of this new section. Not sure if something along `Miscellaneous checks` would be better?~

- I have tested the script itself, but not as part as the CI, because I'm not sure if they're a simple way to redeploy it to test it.

- The script itself behaves somewhat like `diff`: it is expected that no output is produced if no error is found.

### DIFF
--- a/travis-ci/vmtest/run_selftests.sh
+++ b/travis-ci/vmtest/run_selftests.sh
@@ -28,6 +28,12 @@ test_verifier() {
 	travis_fold end test_verifier
 }
 
+bpftool_checks() {
+	travis_fold start bpftool_checks "bpftool checks"
+	./test_bpftool_synctypes.py
+	travis_fold end bpftool_checks
+}
+
 travis_fold end vm_init
 
 configs_path='libbpf/travis-ci/vmtest/configs'


travis-ci/vmtest/run_selftests.sh
```diff
@@ -28,6 +28,12 @@ test_verifier() {
 	travis_fold end test_verifier
 }
 
+test_bpftool() {
+	travis_fold start test_bpftool "Testing bpftool"
```
anakryiko, line 32: bpftools_checks "bpftool checks" ? "check" doesn't mean "run and test", so seems fitting.

Also note that we have libbpf CI  at https://github.com/libbpf/libbpf which uses very similar set of tests, so would be good to have it there as well, I suppose.

qmonnet: Sounds good, I changed to “bpftool checks”.

Yes, I remember you pointed to libbpf's CI too. I was planning to have the current PR merged here first, and to follow up with libbpf/libbpf, but given that the files are so similar I could understand if you prefer to merge both at the same time. So I opened a similar PR at https://github.com/libbpf/libbpf/pull/352.

travis-ci/vmtest/run_selftests.sh
```diff
@@ -28,6 +28,12 @@ test_verifier() {
 	travis_fold end test_verifier
 }
 
+bpftool_checks() {
```
anakryiko, line 31: well, actually, bpftool_checks is not really invoked from anywhere...

qmonnet: Silly me :cry:. Addressed in https://github.com/kernel-patches/vmtest/pull/23, and I'll update the libbpf/libbpf version right now.

PR #27, opened 2021-08-12
Enable CI testing for PRs in vmtest repo

thefallentree:
Adds two optional step in CI workflow that clones bpf-next tree, these steps
only runs out of "kernel-patches/bpf" repo, this will allow testing CI scripts
on vmtest repo PRs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,5 +19,15 @@ jobs:
       - uses: actions/checkout@v2
       - name: install
         run: sudo apt-get update && sudo apt-get install aptitude qemu-kvm zstd binutils-dev elfutils libcap-dev libelf-dev libdw-dev python3-docutils
+      - name: Get bpf-next source
+        if: ${{ github.repository != 'kernel-patches/bpf' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -19,5 +19,15 @@ jobs:
       - uses: actions/checkout@v2
       - name: install
         run: sudo apt-get update && sudo apt-get install aptitude qemu-kvm zstd binutils-dev elfutils libcap-dev libelf-dev libdw-dev python3-docutils
+      - name: Get bpf-next source
+        if: ${{ github.repository != 'kernel-patches/bpf' }}
```
anakryiko, line 23: what's the longer-term solution for this? once we add net/net-next tree, for example, will we keep extending this hard-coded if?

tsipa: This is a good POC. The long term solution would be to keep .gh/wfs/test.yml solely for purpose of testing of CI repo itself and migrate all projects to dedicated folders : https://github.com/tsipa/ci/tree/master/custom-workflows/bpf

PR #29, opened 2021-08-18
Avoid hanging on kernel panics

thefallentree:
using tip from [0], make linux reboot immediately after a kernel panic, and make qemu exit on reboot attempts.

[0] https://unix.stackexchange.com/questions/443017/can-i-make-qemu-exit-with-failure-on-kernel-panic

### DIFF
--- a/travis-ci/vmtest/run.sh
+++ b/travis-ci/vmtest/run.sh
@@ -442,10 +442,10 @@ sudo umount "$mnt"
 
 echo "Starting VM with $(nproc) CPUs..."
 
-qemu-system-x86_64 -nodefaults -display none -serial mon:stdio \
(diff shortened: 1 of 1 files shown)


anakryiko:
Looks good, I guess, but before merging this, let's try implementing this in libbpf CI to see if all the tests are still working properly (would be also nice to simulate panic somehow, but I'm not sure how hard it is to do that as a one-off test).

thefallentree:
> Looks good, I guess, but before merging this, let's try implementing this in libbpf CI to see if all the tests are still working properly (would be also nice to simulate panic somehow, but I'm not sure how hard it is to do that as a one-off test).

the CI in vmtest is working now,  so we can be sure it won't hurt to merge :-D   I can see how I can simulate a crash

PR #33, opened 2021-09-07
Set qemu output buffer to 4K

thefallentree:
see if this will improve runtime


anakryiko:
How will we know if it improves anything?

thefallentree:
no change in run time.

PR #35, opened 2021-09-08
Allow local patch to fail

thefallentree:


anakryiko:
If we silently skip the patch, how will we know about this? Given those patches are temporary, it's probably ok to fail if they don't apply cleanly. This is the problem for Yonghong's most recent patch only, I think it's fine that it is failing.

thefallentree:
> If we silently skip the patch, how will we know about this? Given those patches are temporary, it's probably ok to fail if they don't apply cleanly. This is the problem for Yonghong's most recent patch only, I think it's fine that it is failing.

we will discover that when test fails, right? If the test didn't fail , it is very likely the patch are no longer needed anyway. Not auto-skipping patches would probably making all tests fail after upgrading bpf-next, some one would then have to delete the patch here before CI is working.

anakryiko:
On the other hand we'll have patches laying around and everyone will forget about them. So how will we remember to clean up those *temporary* patches?

But ok, let's try this for now. Can you please at least add some warning message mentioning which patch failed to apply, so that we have that recorded in the log? It's probably a good idea to also record when patch is successfully applied.

thefallentree:
> On the other hand we'll have patches laying around and everyone will forget about them. So how will we remember to clean up those _temporary_ patches?
> 
> But ok, let's try this for now. Can you please at least add some warning message mentioning which patch failed to apply, so that we have that recorded in the log? It's probably a good idea to also record when patch is successfully applied.

patching successfully & failed will both leave a line in the log , as seen here 

image

image

anakryiko:
@thefallentree so this patch is now failing to apply to bpf tree, see https://github.com/kernel-patches/bpf/runs/3557663049?check_suite_focus=true

So I have few requests from that:
1. can we have a per-tree patches, so that we can apply relevant patches to relevant trees only
2. it would be best to use `git am -3` command to apply patches, which will also log any failed hunks right there in the log. This will help to understand what changed and what needs fixing.

PR #36, opened 2021-09-09
Delete local patch

thefallentree:
Do not merge until relevant upstream patch is merged, and CI is green


anakryiko:
Ok, seems like we don't need these patches anymore. Can you please also delete the one for bpf tree? Neither seems to be applied successfully anymore.

Also, it seems like CI logic reports when patch fails to apply, but doesn't report anything when patch successfully applied. Do you think this can be made more obvious?

PR #40, opened 2021-10-01
Dry-run before applying the patch

thefallentree:
this will make sure the patch is always applied cleanly with no reject


anakryiko:
Superseded by #39.

Issue #41, opened 2021-10-03
Change condition on `kernel-patches/bpf` check to detect if Linux sources should be downloaded


qmonnet:
Context: I'm reusing some of the CI infra for a Linux fork on my own GitHub account.

Following commit https://github.com/kernel-patches/vmtest/commit/fa07d13393d0cf5053e9e4b2e396279af940296b, the description of the GitHub workflow has the following snippet to allow the CI to run on the `vmtest` repository.

```yaml
      - name: Get bpf-next source
        if: ${{ github.repository != 'kernel-patches/bpf' }}
        uses: actions/checkout@v2
        with:
          repository: kernel-patches/bpf
          ref: 'bpf-next'
          path: 'linux'
      - name: Move linux source in place
        if: ${{ github.repository != 'kernel-patches/bpf' }}
        run: rsync --exclude .git -av linux/. . && rm -rf linux
```

This makes the Action download kernel sources before compiling and running the checks, if `github.repository != 'kernel-patches/bpf'`.

It is rather easy to copy all files from `vmtest` to another Linux repository and to run the same CI checks from another GitHub account by creating a dedicated PR (e.g. https://github.com/qmonnet/linux/pull/4, the branch was manually created, without the Patchwork parts of the CI), except that the above check will overwrite the Linux sources (and the modifications to test). It's not  hard to remove this check before copying the file and running the action, but it makes the operation less straightforward.

Would it be possible to adjust this check? One solution could be to have `if: ${{ github.repository == 'kernel-patches/vmtest' }}`, although the same issue would arise again if someone tried to run the checks on e.g. a fork of `vmtest`. Alternatively, would it be possible to check whether the repository seems to contain the Linux sources? E.g. by looking at a few known files maybe?

anakryiko:
I'm trying to understand what the actual problem is? You have some local (not committed) changes to the test that get overwritten by that rsync command? Or what exactly is the issue?

qmonnet:
So the CI is presently made to run on two repos:

1. On `kernel-patches/bpf`, by copying all CI files as an additional commit. In that case, the CI does not download Linux sources: it runs on the code committed in the PR.
2. On `kernel-paches/vmtest`, to check that the CI still works as expected on modifications of the CI infra. For that case, we don't copy the CI files, but instead we pull the Linux sources to run on it the new CI files modified in the PR.

[...]

PR #42, opened 2021-10-06
Reverse condition for copying the Linux sources

qmonnet:
Context: https://github.com/kernel-patches/vmtest/issues/41

> kernel-patches/vmtest is an exception here, not the kernel-patches/bpf

We don't want a hardcoded condition on `kernel-patches/bpf` to enable the CI for vmtest, because it makes it harder to reuse the CI on different repos in the future (think net/net-next) or on different forks of bpf-next. Instead, let's copy the sources only when running the GitHub Action on the kernel-patches/vmtest repository.

Next step should be to create a separate .yml workflow description, specific to kernel-patches/vmtest.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,14 +21,14 @@ jobs:
       - name: install
         run: sudo apt-get update && sudo apt-get install aptitude qemu-kvm zstd binutils-dev elfutils libcap-dev libelf-dev libdw-dev python3-docutils
       - name: Get bpf-next source
-        if: ${{ github.repository != 'kernel-patches/bpf' }}
+        if: ${{ github.repository == 'kernel-patches/vmtest' }}
         uses: actions/checkout@v2
         with:
           repository: kernel-patches/bpf
           ref: 'bpf-next'
           path: 'linux'
       - name: Move linux source in place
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -20,15 +20,5 @@ jobs:
...
         run: sudo apt-get update && sudo apt-get install aptitude qemu-kvm zstd binutils-dev elfutils libcap-dev libelf-dev libdw-dev python3-docutils
-      - name: Get bpf-next source
-        if: ${{ github.repository != 'kernel-patches/bpf' }}
-        uses: actions/checkout@v2
-        with:
-          repository: kernel-patches/bpf
-          ref: 'bpf-next'
-          path: 'linux'
-      - name: Move linux source in place
-        if: ${{ github.repository != 'kernel-patches/bpf' }}
```
anakryiko, line 31: we still need this check to make CI runs possible on PRs for kernel-patchs/vmtest. So don't delete it (yet?) until we have a solution for both scenarios. Just invert the condition and hardcode kernel-patches/vmtest repo to keep everything working.

qmonnet: Right, I misunderstood your comment on the other issue. I was a bit surprised that we should entirely remove this step :). I'll adjust the PR.

PR #43, opened 2021-10-19
Enable parallel test execution

thefallentree:


anakryiko:
As discussed offline, let's do this *in addition* to sequential run. That way we have a reliable sequential signal for gating patch application, but also we can get signal on which tests are not working well in parallel mode.

PR #44, opened 2021-11-05
ci: return with non-zero status code if bpftool checks fail

qmonnet:
The bpftool checks work as expected when the CI runs, except that they do not set any error status code for the script on error, which means that the failures are lost among the logs and not reported in a clear way to the reviewers.

Let's account for the status of bpftool checks when exiting from run.sh

[EDIT] Let's also create a summary for the different group tests, at the end of the test logs, under the folded sections. This summary may be extended in the future with a more detailed overview of the tests running in the VM.

**Warning:** Submitted as a PR for discussion and for validation through the CI. Before merging, we should address the current inconsistencies reported by `test_bpftool_synctypes.py`, or we will cause the next CI runs to fail.

### DIFF
--- a/travis-ci/vmtest/run.sh
+++ b/travis-ci/vmtest/run.sh
@@ -431,8 +431,7 @@ sudo chmod 755 "$mnt/etc/rcS.d/S50-run-tests"
 
 poweroff_script="#!/bin/sh
 
-echo travis_fold:start:shutdown
-echo -e '\033[1;33mShutdown\033[0m\n'
+echo -e '$(travis_fold start shutdown Shutdown)'
 
 poweroff"
 echo "${poweroff_script}" | sudo tee "$mnt/etc/rcS.d/S99-poweroff" > /dev/null
--- a/travis-ci/vmtest/run.sh
+++ b/travis-ci/vmtest/run.sh
@@ -378,10 +378,20 @@ LIBBPF_PATH="${REPO_ROOT}" \
 	VMTEST_ROOT="${VMTEST_ROOT}" \
 	VMLINUX_BTF=${vmlinux} ${VMTEST_ROOT}/build_selftests.sh
 
+declare -A test_results
+
 travis_fold start bpftool_checks "Running bpftool checks..."
 if [[ "${KERNEL}" = 'LATEST' ]]; then
-	"${REPO_ROOT}/tools/testing/selftests/bpf/test_bpftool_synctypes.py" && \
-		echo "Consistency checks passed successfully."
+	# "&& true" does not change the return code (it is not executed if the
+	# Python script fails), but it prevents the trap on ERR set at the top
+	# of this file to trigger on failure.
+	"${REPO_ROOT}/tools/testing/selftests/bpf/test_bpftool_synctypes.py" && true
+	test_results["bpftool"]=$?
+	if [[ ${test_results["bpftool"]} -eq 0 ]]; then
+		echo "::notice title=bpftool_checks::bpftool checks passed successfully."
+	else
+		echo "::error title=bpftool_checks::bpftool checks returned ${test_results["bpftool"]}."
+	fi
 else
 	echo "Consistency checks skipped."
 fi
@@ -466,5 +476,22 @@ fi
 sudo umount "$mnt"
 
 travis_fold end shutdown
+
+test_results["vm_tests"]=$exitstatus
+
+# Final summary - Don't use a fold, keep it visible
+echo -e "\033[1;33mTest Results:\033[0m"
+for testgroup in ${!test_results[@]}; do
+	# Print final result for each group of tests
+	if [[ ${test_results[$testgroup]} -eq 0 ]]; then
+		printf "%20s: \033[1;32mPASS\033[0m\n" $testgroup
+	else
+		printf "%20s: \033[1;31mFAIL\033[0m\n" $testgroup
+	fi
+	# Make exitstatus > 0 if at least one test group has failed
+	if [[ ${test_results[$testgroup]} -ne 0 ]]; then
+		exitstatus=1
+	fi
+done
 
 exit "$exitstatus"


qmonnet:
As expected the CI run fails, because at the moment `test_bpftool_synctypes.py` reports some inconsistencies.

I note that it's not easy to identify where the failure happened: the interface for the GitHub action opens the `test_verifier - Testing test_verifier` subsection (and not the one of the bpftool checks), presumably because it was the last to run when the error code was raised? I'm not sure how to improve that. I suppose one option could be to `exit 1` directly after the bpftool checks when they fail, but I believe this would prevent the rest of the checks to run, and I don't think this is desirable?

anakryiko:
Maybe let's add extra check that bpftool_exitstatus is non-zero, emit log message and exit 1. We should do that for each separate check (test_progs, test_progs-no_alu32, test_maps, test_Verifier, bpftool checks, etc). So that we run all of them even if some of them fail. But at the end give some pointers as to what exactly failed.

This setup would allow each section to be collapsed by default and just have diagnostic summary at the very end for anyone who's doing the investigation. Try doing that for bpftool and see how it works.

The inconsistency has to be fixed upstream before this lands as well, of course.

qmonnet:
> The inconsistency has to be fixed upstream before this lands as well, of course.

Sure, but in the meantime it's convenient for testing this PR :)
Ok I'll look into it, thanks for the suggestions.

qmonnet:
Please let me know what you think of the current version.

travis-ci/vmtest/run.sh
```diff
@@ -378,10 +378,17 @@ LIBBPF_PATH="${REPO_ROOT}" \
 	VMTEST_ROOT="${VMTEST_ROOT}" \
 	VMLINUX_BTF=${vmlinux} ${VMTEST_ROOT}/build_selftests.sh
 
+declare -A tests_results
+
 travis_fold start bpftool_checks "Running bpftool checks..."
 if [[ "${KERNEL}" = 'LATEST' ]]; then
-	"${REPO_ROOT}/tools/testing/selftests/bpf/test_bpftool_synctypes.py" && \
-		echo "Consistency checks passed successfully."
+	"${REPO_ROOT}/tools/testing/selftests/bpf/test_bpftool_synctypes.py" && true
```
anakryiko, line 385: does `&& true` changes anything? Doesn't it overrides the exit code? I suspect I'm missing some subtle behavior here, can you please explain?

travis-ci/vmtest/run.sh
```diff
@@ -378,10 +378,17 @@ LIBBPF_PATH="${REPO_ROOT}" \
...
+	tests_results["bpftool"]=$?
+	if [[ ${tests_results["bpftool"]} -eq 0 ]]; then
+		echo "::notice title=bpftool_checks::bpftool checks passed successfully."
+	else
+		echo "::error title=bpftool_checks::bpftool checks returned with ${tests_results["bpftool"]}."
```
anakryiko, line 390: s/returned with/returned/ ? seems to be reading more naturally

travis-ci/vmtest/run.sh
```diff
@@ -468,4 +474,19 @@ sudo umount "$mnt"
...
+echo -e "\033[1;33mTest Results:\033[0m"
+for testgroup in ${!tests_results[@]}; do
+	# Print final result for each group of tests
+	if [[ ${tests_results[$testgroup]} -eq 0 ]]; then
+		printf "%20s: \033[1;32mPASS\033[0m\n" $testgroup
+	else
+		printf "%20s: \033[1;31mFAIL\033[0m\n" $testgroup
+	fi
+	# Make exitstatus > 0 if at least one test group has failed
+	exitstatus=$((exitstatus + tests_results[$testgroup]))
```
anakryiko, line 489: adding exitstatuses seems error prone and a bit confusing, maybe if + exitstatus = 1?

travis-ci/vmtest/run.sh
```diff
@@ -378,10 +378,17 @@ LIBBPF_PATH="${REPO_ROOT}" \
 	VMTEST_ROOT="${VMTEST_ROOT}" \
 	VMLINUX_BTF=${vmlinux} ${VMTEST_ROOT}/build_selftests.sh
 
+declare -A tests_results
```
anakryiko, line 381: nit: test_results (one result per test, so in English should be "test results")

anakryiko (commented):
Looks great, just few small nits. Do we do the same check in libbpf CI? If yes, I'll ask you to submit another PR there because we still hasn't consolidated BPF CI scripts :(

travis-ci/vmtest/run.sh
qmonnet: It does not change the exit code, because if the Python script fails, the `echo true` is never evaluated.

But it does have an effect: Because we have a `trap 'exit 2' ERR` at the beginning of the file, `run.sh` would exit if the Python script failed. Adding the `&& true` prevents that. Here's what the man page for bash says:

> The `ERR` trap is not executed if the failed command is [...] part of a command executed in a `&&` or `||` list except the command following the final `&&` or `||`, [...].

I'll add a comment to make it clear.

travis-ci/vmtest/run.sh
qmonnet: But there may be several tests per results, depending on how we consider things (technically, one result per group of test), and that probably got me confused. `test_results` does sound better, I'll adjust.

qmonnet:
> Do we do the same check in libbpf CI? If yes, I'll ask you to submit another PR there because we still hasn't consolidated BPF CI scripts :(

We do, and sure, I can open a similar PR for the libbpf repo. Let's finalise this one here first? Then I need to fix the bpftool inconsistencies in bpf-next anyway, so I can create the twin PR in libbpf in parallel.

<details><summary>Changes from previous version</summary>

```diff
diff --git a/travis-ci/vmtest/run.sh b/travis-ci/vmtest/run.sh
index aa6dfd02b60b..4a2f81d65d22 100755
--- a/travis-ci/vmtest/run.sh
+++ b/travis-ci/vmtest/run.sh
@@ -378,16 +378,19 @@ LIBBPF_PATH="${REPO_ROOT}" \
 	VMTEST_ROOT="${VMTEST_ROOT}" \
 	VMLINUX_BTF=${vmlinux} ${VMTEST_ROOT}/build_selftests.sh
 
-declare -A tests_results
+declare -A test_results
 
 travis_fold start bpftool_checks "Running bpftool checks..."
 if [[ "${KERNEL}" = 'LATEST' ]]; then
+	# "&& true" does not change the return code (it is not executed if the
+	# Python script fails), but it prevents the trap on ERR set at the top
+	# of this file to trigger on failure.
 	"${REPO_ROOT}/tools/testing/selftests/bpf/test_bpftool_synctypes.py" && true
-	tests_results["bpftool"]=$?
-	if [[ ${tests_results["bpftool"]} -eq 0 ]]; then
+	test_results["bpftool"]=$?
+	if [[ ${test_results["bpftool"]} -eq 0 ]]; then
 		echo "::notice title=bpftool_checks::bpftool checks passed successfully."
 	else
-		echo "::error title=bpftool_checks::bpftool checks returned with ${tests_results["bpftool"]}."
+		echo "::error title=bpftool_checks::bpftool checks returned ${test_results["bpftool"]}."
 	fi
 else
 	echo "Consistency checks skipped."
@@ -474,19 +477,21 @@ sudo umount "$mnt"
 
 travis_fold end shutdown
 
-tests_results["vm_tests"]=$exitstatus
+test_results["vm_tests"]=$exitstatus
 
 # Final summary - Don't use a fold, keep it visible
 echo -e "\033[1;33mTest Results:\033[0m"
-for testgroup in ${!tests_results[@]}; do
+for testgroup in ${!test_results[@]}; do
 	# Print final result for each group of tests
-	if [[ ${tests_results[$testgroup]} -eq 0 ]]; then
+	if [[ ${test_results[$testgroup]} -eq 0 ]]; then
 		printf "%20s: \033[1;32mPASS\033[0m\n" $testgroup
 	else
 		printf "%20s: \033[1;31mFAIL\033[0m\n" $testgroup
 	fi
 	# Make exitstatus > 0 if at least one test group has failed
-	exitstatus=$((exitstatus + tests_results[$testgroup]))
+	if [[ $test_results[$testgroup] -ne 0 ]]; then
+		exitstatus=1
+	fi
 done
 
 exit "$exitstatus"
```
</details>

anakryiko:
> > Do we do the same check in libbpf CI? If yes, I'll ask you to submit another PR there because we still hasn't consolidated BPF CI scripts :(
> 
> We do, and sure, I can open a similar PR for the libbpf repo. Let's finalise this one here first? Then I need to fix the bpftool inconsistencies in bpf-next anyway, so I can create the twin PR in libbpf in parallel.

Yeah, sure. Changes look good, I'm just waiting for the test run to finish to see how the "::error" / "::notice" stuff looks like. It's the first time I see this.

qmonnet:
Sure, the reference is [here](https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-an-error-message) in case you want to have a look.

qmonnet:
Not sure what I did wrong, but we lost the result from bpftool tests in the summary :(. I'll look into it.

anakryiko:
Ok, cool, it works very nicely. Let's fix the issue upstream and then land this. Then add this to libbpf CI. Thanks!

PR #47, opened 2021-11-24
Split results per group of selftests in final summary

qmonnet:
As a follow-up to the recent introduction of a “final summary” of the selftest results in the CI Action logs, make the different groups of selftests (`test_progs`, `test_progs-no_alu32`, `test_maps`, `test_verifier`) print their results individually.

To do:
- [ ] Validate that the changes work as intended
- [ ] Twin PR for libbpf

### DIFF
--- a/travis-ci/vmtest/run.sh
+++ b/travis-ci/vmtest/run.sh
@@ -276,10 +276,9 @@ if [[ $SKIPIMG -eq 0 && ! -v ROOTFSVERSION ]]; then
 	ROOTFSVERSION="$(newest_rootfs_version)"
 fi
 
-echo "Kernel release: $KERNELRELEASE" >&2
-echo
-
 travis_fold start vmlinux_setup "Preparing Linux image"
+
+echo "Kernel release: $KERNELRELEASE" >&2
 
 if (( SKIPIMG )); then
 	echo "Not extracting root filesystem" >&2
@@ -429,20 +428,23 @@ dmesg -c
 
 echo 'Running setup commands'
 %s
-set +e; %s; exitstatus=\$?; set -e
+set +e;
+%s; exitstatus=\$?;
+echo -e '$(travis_fold start collect_status "Collect status and dmesg")'
+set -e
 # If setup command did not write its exit status to /exitstatus, do it now
 if ! grep -qv bpftool /exitstatus; then
 	echo setup_cmd:\$exitstatus >> /exitstatus
 fi
-dmesg > /dmesg" "${setup_envvars}" "${setup_cmd}")
+dmesg > /dmesg
+echo -e '$(travis_fold end collect_status)'
+echo -e '$(travis_fold start shutdown Shutdown)'" "${setup_envvars}" "${setup_cmd}")
 fi
 
 echo "${setup_script}" | sudo tee "$mnt/etc/rcS.d/S50-run-tests" > /dev/null
 sudo chmod 755 "$mnt/etc/rcS.d/S50-run-tests"
 
 poweroff_script="#!/bin/sh
-
-echo -e '$(travis_fold start shutdown Shutdown)'
 
 poweroff"
 echo "${poweroff_script}" | sudo tee "$mnt/etc/rcS.d/S99-poweroff" > /dev/null
--- a/travis-ci/vmtest/run.sh
+++ b/travis-ci/vmtest/run.sh
@@ -372,13 +372,13 @@ else
 	sudo chmod 644 "$vmlinux"
 fi
 
+sudo truncate -s 0 "$mnt/exitstatus"
+
 travis_fold end vmlinux_setup
 
 LIBBPF_PATH="${REPO_ROOT}" \
 	VMTEST_ROOT="${VMTEST_ROOT}" \
 	VMLINUX_BTF=${vmlinux} ${VMTEST_ROOT}/build_selftests.sh
-
-declare -A test_results
 
 travis_fold start bpftool_checks "Running bpftool checks..."
 if [[ "${KERNEL}" = 'LATEST' ]]; then
@@ -386,14 +386,15 @@ if [[ "${KERNEL}" = 'LATEST' ]]; then
 	# Python script fails), but it prevents the trap on ERR set at the top
 	# of this file to trigger on failure.
 	"${REPO_ROOT}/tools/testing/selftests/bpf/test_bpftool_synctypes.py" && true
-	test_results["bpftool"]=$?
-	if [[ ${test_results["bpftool"]} -eq 0 ]]; then
+	exitstatus=$?
+	sudo bash -c "echo 'bpftool:${exitstatus}' > '$mnt/exitstatus'"
+	if [[ "$exitstatus" -eq 0 ]]; then
 		print_notice bpftool_checks "bpftool checks passed successfully."
 	else
-		print_error bpftool_checks "bpftool checks returned ${test_results["bpftool"]}."
+		print_error bpftool_checks "bpftool checks returned ${exitstatus}."
 	fi
 else
-	echo "Consistency checks skipped."
+	echo "bpftool checks skipped."
 fi
 travis_fold end bpftool_checks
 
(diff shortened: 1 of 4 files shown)


travis-ci/vmtest/run_selftests.sh
```diff
@@ -4,27 +4,41 @@ set -euo pipefail
 
 source $(cd $(dirname $0) && pwd)/helpers.sh
 
+STATUS_FILE=/exitstatus
+
 test_progs() {
 	if [[ "${KERNEL}" != '4.9.0' ]]; then
 		travis_fold start test_progs "Testing test_progs"
+		set +e
```
anakryiko, line 12: you don't really need to set/reset +e, you should be good with adding `|| true` to prevent the shell from erroring out

travis-ci/vmtest/run.sh
```diff
@@ -426,11 +426,12 @@ if [[ ! -z SETUPCMD ]]; then
 	setup_script=$(printf "#!/bin/sh
 set -eux
 dmesg -c
+> /exitstatus
```
anakryiko, line 429: is this to create a file? quite unusual, I usually do `touch <file>` for that

travis-ci/vmtest/run.sh
```diff
@@ -457,7 +458,13 @@ qemu-system-x86_64 -nodefaults -display none -serial mon:stdio -no-reboot \
 	-kernel "$vmlinuz" -append "root=/dev/vda rw console=ttyS0,115200 kernel.panic=-1 $APPEND"
 
 sudo mount -o loop "$IMG" "$mnt"
-if exitstatus="$(cat "$mnt/exitstatus" 2>/dev/null)"; then
+
+while read l; do
+	test_results[${l%:*}]=${l#*:}
```
anakryiko, line 463: it would be nice to preserve the order of execution, I think this bash's map actually loses it. Is there any way to fix that?

anakryiko (commented):
This is really nice, thanks! Few nits in the comments, but I also have few questions/asks:

1. Is it possible to have a link back to the end of the section that corresponds to test step. When something fails, it would be nice to just click "FAILED" and go to the relevant section of the log. I think Github Actions had some anchors that could be used for this, no?
2. Can you please see if we can hide all the irrelevant parts of the logs into sections. E.g., currently there is:
```
 + exitstatus=0
+ set -e
+ echo setup_cmd:0
+ dmesg
+ chmod 644 /exitstatus
+ for path in /etc/rcS.d/S*
+ '[' -x /etc/rcS.d/S99-poweroff ']'
+ /etc/rcS.d/S99-poweroff
```

that could go into shutdown. There is also a bunch of garbage before the vm setup, I think.

Thanks!

travis-ci/vmtest/run_selftests.sh
qmonnet: But that wouldn't allow me to get the correct return code with `$?` on the following line, I believe?

travis-ci/vmtest/run.sh
qmonnet: No, this is to truncate it, in the (admittedly unlikely) case it's already existing. I don't see it so often, although I thought it was rather admitted as a classic way to truncate a file. I can so some research or change to `printf '' > /exitfile` if you prefer. Or add a comment.

travis-ci/vmtest/run.sh
qmonnet: There's always a way, I guess, but that will probably require a bit more processing. We may have to drop the associative array and find something else, to store both the result and the order of execution.

qmonnet:
I need to look at the anchors and at stashing these logs into existing subsections, I'll update the PR. Thanks for the review and feedback!

travis-ci/vmtest/run_selftests.sh
anakryiko: It should be `&& true`, sorry. We do that already: https://github.com/kernel-patches/vmtest/blob/master/travis-ci/vmtest/run.sh#L385-L389

travis-ci/vmtest/run.sh
anakryiko: Oh, truncation. Yeah, it's fine, but let's add a brief comment.

travis-ci/vmtest/run.sh
anakryiko: Yeah, given we have sequential file, we can do just that, right? It could be simpler to have "test_name 0" format to do readline later.

travis-ci/vmtest/run_selftests.sh
qmonnet: Right, should have remembered that :facepalm:. The `||` put me off. Ok will do, it will be more readable.

travis-ci/vmtest/run.sh
qmonnet: Yep, we'll just need to deal with the bpftool results that we get first but that isn't written to the file. I'll sort this out.

anakryiko: Why not write bpftool's results to the same file in the same format?..

qmonnet: I can write to the file, I think, since we access it (read it) from outside the VM. I was a bit wary of that because the setup commands could have erased bpftool's results; but that's probably too much precaution. It also makes it slightly more difficult to know if the setup command has written its result (not an issue in the CI I think, this would be in case we run the script independently with a custom `--setup-cmd`), but it's just a matter of grepping out for `bpftool`.

I updated the PR to try and write all to the `/exitstatus` file, let's see how it goes.

qmonnet:
I haven't found any documentation for the anchors, do you have a link or an example around by any chance?

anakryiko:
> I haven't found any documentation for the anchors, do you have a link or an example around by any chance?

Never mind, I've mistakenly assumed that https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-an-error-message could help, but it can only reference files in the repo, no the log itself. Too bad.

qmonnet:
@anakryiko can you have a look at the current output, please? The summary seems to be working (I'll test by forcing a test to fail next).

I still have some logs outside of subsections, because
- I cannot end a subsection right after the `setup_cmd`, given that I need to extract `$?` first
- When the section markers are in scripts called with `set -eux` (like `/etc/rcS.d/S50-run-tests` in the VM), the command to open/close the log section displays first with a `+` prefix due to the `-x` option, and appears out of the sections. Possible workarounds: 1) remove `-x` from script (also addresses the first bullet point), 2) split script into two parts, print sections in between, 3) keep a few log lines outside of the sections. I think 1) would give a cleaner output but we lose some info from that script (e.g. the exact `setup_cmd` which runs; unless we `echo` it).
- I'm not sure why the `Kernel release: 5.15.0-g563a45dae677-dirty` shows up outside of a section. It should be printed after the opening of the `vmlinux_setup` section. It's supposed to be printed to stderr though, so I wonder if there might be a race between stdout and stderr? Probably doesn't matter too much.

anakryiko:
> @anakryiko can you have a look at the current output, please? The summary seems to be working (I'll test by forcing a test to fail next).
> 

Looks great, applying. Thanks for improving this!

> I still have some logs outside of subsections, because
> 
>     * I cannot end a subsection right after the `setup_cmd`, given that I need to extract `$?` first
> 
>     * When the section markers are in scripts called with `set -eux` (like `/etc/rcS.d/S50-run-tests` in the VM), the command to open/close the log section displays first with a `+` prefix due to the `-x` option, and appears out of the sections. Possible workarounds: 1) remove `-x` from script (also addresses the first bullet point), 2) split script into two parts, print sections in between, 3) keep a few log lines outside of the sections. I think 1) would give a cleaner output but we lose some info from that script (e.g. the exact `setup_cmd` which runs; unless we `echo` it).

I think we can remove -x easily with no harm.

> 
>     * I'm not sure why the `Kernel release: 5.15.0-g563a45dae677-dirty` shows up outside of a section. It should be printed after the opening of the `vmlinux_setup` section. It's supposed to be printed to stderr though, so I wonder if there might be a race between stdout and stderr? Probably doesn't matter too much.

Probably because of stderr. But it's totally fine and kind of nice, so don't bother fixing this.

PR #48, opened 2021-11-30
ci: fix test on exitstatus

qmonnet:
The check on the `$exitstatus` collected from the `/exitstatus` file was changed in a recent commit, and is not correct. Let's fix it. We consider that we failed to read the exit status if we get something other than an integer. This might happen if awk fails because it cannot access the file, for example.

### DIFF
--- a/travis-ci/vmtest/run.sh
+++ b/travis-ci/vmtest/run.sh
@@ -465,10 +465,11 @@ sudo mount -o loop "$IMG" "$mnt"
 exitstatus=$(sudo awk --field-separator ':' \
 	'BEGIN { s=0 } { if ($2) {s=1} } END { print s }' "$mnt/exitstatus")
 
-if [[ "$exitstatus" -ne 0 ]]; then
+if [[ "$exitstatus" =~ ^[0-9]+$ ]]; then
 	printf '\nTests exit status: %s\n' "$exitstatus" >&2
 else
-	printf '\nCould not read tests exit status\n' >&2
+	printf '\nCould not read tests exit status ("%s")\n' "$exitstatus" >&2
+	exitstatus=1
 fi
 if [ -f "$mnt/dmesg" ]; then
         ! grep "RIP" "$mnt/dmesg" -C 20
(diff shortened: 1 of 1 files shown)


travis-ci/vmtest/run.sh
```diff
@@ -465,10 +465,11 @@ sudo mount -o loop "$IMG" "$mnt"
 exitstatus=$(sudo awk --field-separator ':' \
 	'BEGIN { s=0 } { if ($2) {s=1} } END { print s }' "$mnt/exitstatus")
 
-if [[ "$exitstatus" -ne 0 ]]; then
+if [[ "$exitstatus" =~ '^[0-9]+$' ]]; then
 	printf '\nTests exit status: %s\n' "$exitstatus" >&2
 else
 	printf '\nCould not read tests exit status\n' >&2
```
anakryiko, line 471: might be useful to log "$exitstatus" if it's not a number?

anakryiko (commented):
You forgot to add your Signed-off-by, btw.

qmonnet:
> You forgot to add your Signed-off-by, btw.

I didn't forget. I thought it wasn't a requirement for this repo and haven't been using it so far (19 commits out of the 89 have a tag, all but one of which are yours). I can start using it if you'd like.

anakryiko:
> > You forgot to add your Signed-off-by, btw.
> 
> I didn't forget. I thought it wasn't a requirement for this repo and haven't been using it so far (19 commits out of the 89 have a tag, all but one of which are yours). I can start using it if you'd like.

I saw Fixes: tag, so assumed we go with kernel conventions. Generally I tried to steer people to include their SOBs for libbpf, libbpf-rs and libbpf-bootstrap repos to keep it a bit more consistent across. Not a hard requirement, but also not much harm to add it, no?

qmonnet:
> Generally I tried to steer people to include their SOBs for libbpf, libbpf-rs and libbpf-bootstrap repos to keep it a bit more consistent across. Not a hard requirement, but also not much harm to add it, no?

When I sign-off, I take an additional step to engage my responsibility for the commit (I'm not sure what the legal value is, but there is at least the moral aspect). I don't gain much personally by doing so, so I don't usually do it for a project that does not use the tags. It doesn't mean that I bring less care to the code or to my right to send the patch, of course. And I'm entirely fine signing off for repos who use it. Note that this is not necessarily the best reasoning and I suppose one could have a different vision, and prefer to sign-off all the time to show that they're confident and willing to commit themselves.

Then from a maintainer's/project's point of view, it's the opposite: better have the certificate as some kind of “insurance” that the code is legit (not sure that's the best terms, but I suppose you get the idea), so I totally understand if on your side you prefer to have the tags.

All of this to say: got it, I'll add the tags for those projects from now on. Maybe you should encourage people to tag their commits in the respective READMEs?

PR #52, opened 2021-12-20
Move patch-kernel to libbpf/ci

thefallentree:


anakryiko:
Similar to kernel configs, should the set of patches be part of specific repo (vmtest vs libbpf)?

anakryiko (approved):

PR #55, opened 2022-01-05
use "-j" for test_progs if core count is over than 2

thefallentree:
This will make CI run faster on new self-hosted runner with more cpus


anakryiko:
I'm worried that parallel mode is still going to spuriously fail and we'll have to revert this change. I thought we are going to have parallel run in addition to serial run to inform us on which test are not parallel safe yet. Please keep track of CI runs stability for the time being.

PR #57, opened 2022-01-06
[do not merge] debug test parallelization

thefallentree:
leave 2 cpus, this may fix the softlockup issue

### DIFF
--- a/.github/actions/vmtest/action.yml
+++ b/.github/actions/vmtest/action.yml
@@ -23,7 +23,7 @@ runs:
         arch: 'x86_64'
         kernel-root: '.'
     - name: Run selftests
-      uses: libbpf/ci/run-qemu@thefallentree
+      uses: libbpf/ci/run-qemu@master
       with:
         arch: 'x86_64'
         img: '/tmp/root.img'
--- a/.github/actions/vmtest/action.yml
+++ b/.github/actions/vmtest/action.yml
@@ -23,7 +23,7 @@ runs:
         arch: 'x86_64'
         kernel-root: '.'
     - name: Run selftests
-      uses: libbpf/ci/run-qemu@master
+      uses: libbpf/ci/run-qemu@thefallentree
       with:
         arch: 'x86_64'
         img: '/tmp/root.img'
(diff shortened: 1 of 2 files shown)


thefallentree:
> 

that's actually not true, the test run 2 times (test_prog and test_prog_noalu32) , each one about 5minutes , so it contributes 10minutes to total run time -> 45 minute vs 20 minutes . yeah it is a experiment to see if we can leave it on, because during test execution the whole machine is sitting idle, wasting money

anakryiko:
> contributes 10minutes to total run time -> 45 minute vs 20 minutes

How much speed up do we get from parallelizing test_progs and test_progs-no_alu32? It goes down from 10 to what number of minutes?

Can we do this experiment without affecting CI normal functioning? We do rely on stability of CI, so if we keep breaking it back and forth it just reduces the trustworthiness of the whole BPF CI. Can you add parallelized run as a separate optional step?

thefallentree:
> > contributes 10minutes to total run time -> 45 minute vs 20 minutes
> 
> How much speed up do we get from parallelizing test_progs and test_progs-no_alu32? It goes down from 10 to what number of minutes?

It's about 5 minutes to 1 minutes


> 
> Can we do this experiment without affecting CI normal functioning? We do rely on stability of CI, so if we keep breaking it back and forth it just reduces the trustworthiness of the whole BPF CI. 

That's what this PR is for, this repo is enabled with big runners and I'm trying to test if this works.


> Can you add parallelized run as a separate optional step?

Yes, but it is a bigger change require a bit more work

anakryiko:
> > > contributes 10minutes to total run time -> 45 minute vs 20 minutes
> > 
> > 
> > How much speed up do we get from parallelizing test_progs and test_progs-no_alu32? It goes down from 10 to what number of minutes?
> 
> It's about 5 minutes to 1 minutes

Great, but let's keep stability as a first priority.

> 
> > Can we do this experiment without affecting CI normal functioning? We do rely on stability of CI, so if we keep breaking it back and forth it just reduces the trustworthiness of the whole BPF CI.
> 
> That's what this PR is for, this repo is enabled with big runners and I'm trying to test if this works.

Ok, it wasn't clear that it's for testing. I see you marked it as draft now, thanks.

> 
> > Can you add parallelized run as a separate optional step?
> 
> Yes, but it is a bigger change require a bit more work

See above about the stability, we rely on this system, so when it's not working properly it's disruptive. So I understand that it might be more worked to do things in more stable way, but I think it's worth it.

PR #65, opened 2022-03-17
Add support for testing on s390x ibm setup

mykola-lysenko:


.github/actions/vmtest/build.sh
```diff
@@ -2,13 +2,16 @@
 
 set -euo pipefail
 
+ARCH="$1"
```
anakryiko, line 5: in run_selftests.sh we just do `ARCH=$(uname -m)`, why can't we do the same here? It seems simpler, if it works reliably.

anakryiko (approved):
LGTM, see the bit about keeping build.sh simpler, but it's very minor.

.github/actions/vmtest/build.sh
mykola-lysenko: we in run_selftests.sh we do ARCH=$(uname -m) because there is not way to pass this information from GitHub Action. On the other hand, in the GH Action we pass this information around as a parameter. I prefer to pass it as a parameter in this case, as it will be better for cross-compilation if it ever happens.

I am not too strong about this, happy to change if you prefer

mykola-lysenko:
> 

It is xdp_do_redirect test. Happy to share more information if needed

iii-i:
Hi, this looks great, thanks! I'm currently wrapping up the serval/jitterbug work, and starting looking into trampolines. This should reduce the list significantly.

PR #77, opened 2022-04-26
ci: enable CONFIG_MPTCP=y for all architectures

anakryiko:
Enable CONFIG_MPTCP in Kconfig for x86-64 and s390x.

Cc: Daniel Borkmann <daniel@iogearbox.net>
Signed-off-by: Andrii Nakryiko <andrii@kernel.org>


travis-ci/vmtest/configs/config-latest.x86_64
```diff
@@ -961,7 +961,7 @@ CONFIG_IPV6_SEG6_LWTUNNEL=y
 CONFIG_IPV6_SEG6_BPF=y
 # CONFIG_IPV6_RPL_LWTUNNEL is not set
 CONFIG_NETLABEL=y
-# CONFIG_MPTCP is not set
+CONFIG_MPTCP=y
```
matttbe, line 964: Because `CONFIG_IPV6` is also enabled above, you might want to enable `CONFIG_MPTCP_IPV6` as well. I guess you don't need to explicitly add it as it should be enabled automatically with an `olddefconfig` but I prefer to mention it just in case :)

Same in `s390x` config.

(also, `CONFIG_INET_MPTCP_DIAG` will be selected too but no need to worry about this one)

matttbe (commented):
Thank you for this modification!

Just a small comment, just in case, probably fine like that!

PR #82, opened 2022-05-12
ci: add LLVM support

mykola-lysenko:

### DIFF
--- a/.github/actions/vmtest/build_selftests.sh
+++ b/.github/actions/vmtest/build_selftests.sh
@@ -6,9 +6,19 @@ THISDIR="$(cd $(dirname $0) && pwd)"
 
 source "${THISDIR}"/helpers.sh
 
-travis_fold start prepare_selftests "Building selftests"
+TOOLCHAIN="$1"
+TOOLCHAIN_NAME="$(echo $TOOLCHAIN | cut -d '-' -f 1)"
+TOOLCHAIN_VERSION="$(echo $TOOLCHAIN | cut -d '-' -f 2)"
 
+if [ "$TOOLCHAIN_NAME" == "llvm" ]; then
+export LLVM="-$TOOLCHAIN_VERSION"
+LLVM_VER=$TOOLCHAIN_VERSION
+else
 LLVM_VER=15
+fi
+
+travis_fold start prepare_selftests "Building selftests with $TOOLCHAIN"
+
 LIBBPF_PATH="${REPO_ROOT}"
 
 PREPARE_SELFTESTS_SCRIPT=${THISDIR}/prepare_selftests-${KERNEL}.sh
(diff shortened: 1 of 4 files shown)


.github/actions/vmtest/build_selftests.sh
```diff
@@ -6,9 +6,19 @@ THISDIR="$(cd $(dirname $0) && pwd)"
 
 source "${THISDIR}"/helpers.sh
 
-travis_fold start prepare_selftests "Building selftests"
+TOOLCHAIN="$1"
+TOOLCHAIN_NAME="$(echo $TOOLCHAIN | cut -d '-' -f 1)"
+TOOLCHAIN_VERSION="$(echo $TOOLCHAIN | cut -d '-' -f 2)"
 
+if [ "$TOOLCHAIN_NAME" == "llvm" ]; then
+export LLVM="-$TOOLCHAIN_VERSION"
```
anakryiko, line 14: isn't it always LLVM=1 ?

anakryiko: btw, @yonghong-song, FYI, we'll soon start detecting all these LLVM-specific issues early on. Please take a look at this as well.

anakryiko (commented):
looks good (not sure about LLVM=15, though), but let's wait for my fix to land first

.github/actions/vmtest/build_selftests.sh
mykola-lysenko: LLVM toolchain we install in CI has version suffix (currently it is 15).

Linux Kernel build system handles this gracefully: https://github.com/torvalds/linux/blob/b4a5ea09b29371c2e6a10783faa3593428404343/tools/scripts/Makefile.include#L54-L73

I do not like however that it still defaults to clang without version suffix: https://github.com/torvalds/linux/blob/b4a5ea09b29371c2e6a10783faa3593428404343/tools/scripts/Makefile.include#L87-L92. This functionality is used when we build bpf c code in selftests (https://github.com/kernel-patches/vmtest/pull/82/files#diff-c34c251202780f09ab91ab0021a7190227e9a48d8063b004a644fe65a365125fR37-R39).

I am thinking about sending a patch to Linux build system to properly initialize CLANG variable when LLVM variable is set and has a version suffix provided

mykola-lysenko: Some more details in https://www.kernel.org/doc/html/latest/kbuild/llvm.html

anakryiko: nice, didn't know about it.

anakryiko (approved):

.github/actions/vmtest/build_selftests.sh
yonghong-song: Looks like this llvm suffix thing is just extended two months ago:
  https://lore.kernel.org/all/20200317215515.226917-1-ndesaulniers@google.com/
the llvm prefix thing is extended earlier.

I don't have a strong opinion on this as I always use the default path 'clang' only. Whatever it is best for CI is also okay for me.

PR #96, opened 2022-07-18
Clean up Travis remnants

danielocfb:
This pull request proposes the removal/cleanup of a bunch of Travis remnants.

### DIFF
--- a/travis-ci/vmtest/helpers.sh
+++ b/travis-ci/vmtest/helpers.sh
@@ -4,23 +4,15 @@
 foldable() {
   local YELLOW='\033[1;33m'
   local NOCOLOR='\033[0m'
-  if [ -z ${GITHUB_WORKFLOW+x} ]; then
-    echo travis_fold:$1:$2
+  if [ $1 = "start" ]; then
+    line="::group::$2"
     if [ ! -z "${3:-}" ]; then
-      echo -e "${YELLOW}$3${NOCOLOR}"
+      line="$line - ${YELLOW}$3${NOCOLOR}"
     fi
-    echo
   else
-    if [ $1 = "start" ]; then
-      line="::group::$2"
-      if [ ! -z "${3:-}" ]; then
-        line="$line - ${YELLOW}$3${NOCOLOR}"
-      fi
-    else
-      line="::endgroup::"
-    fi
-    echo -e "$line"
+    line="::endgroup::"
   fi
+  echo -e "$line"
 }
 
 __print() {
--- a/travis-ci/vmtest/helpers.sh
+++ b/travis-ci/vmtest/helpers.sh
@@ -1,7 +1,7 @@
 # $1 - start or end
 # $2 - fold identifier, no spaces
 # $3 - fold section description
-travis_fold() {
+foldable() {
   local YELLOW='\033[1;33m'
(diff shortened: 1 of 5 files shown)


travis-ci/vmtest/helpers.sh
```diff
@@ -1,26 +1,18 @@
...
-  if [ -z ${GITHUB_WORKFLOW+x} ]; then
-    echo travis_fold:$1:$2
+  if [ $1 = "start" ]; then
+    line="::group::$2"
     if [ ! -z "${3:-}" ]; then
-      echo -e "${YELLOW}$3${NOCOLOR}"
+      line="$line - ${YELLOW}$3${NOCOLOR}"
     fi
-    echo
   else
```
mykola-lysenko, line 12: can be done as a separate PR

I wonder if it will make sense to either explicitly check for "end" in here or have two separate functions to avoid if and need to pass one of the parameters

mykola-lysenko (approved):
Thanks Daniel!

Added a comment, although it can be addressed a separate PR

Another possible improvement, not sure if this is possible in bash, but can we wrap lists of bash commands with one "foldable" function call? If yes, this will invalidate my other comment

travis-ci/vmtest/helpers.sh
mykola-lysenko: We will also need to backport this to libbpf/ci code or merge them somehow

danielocfb: Thanks for taking a look.

> can be done as a separate PR

It's a separate commit. What would be the benefit of having a separate pull request?

> I wonder if it will make sense to either explicitly check for "end" in here or have two separate functions to avoid if and need to pass first parameter

It has nothing to do with this pull request's intention, though? I am sure there was a reason it was written this way and I don't see a good reason to change it, to be honest.

danielocfb: > We will also need to backport this to libbpf/ci code or merge them somehow

It has already been merged there a while back: https://github.com/libbpf/ci/pull/18

danielocfb:
> Another possible improvement, not sure if this is possible in bash, but can we wrap lists of bash commands with one "foldable" function call? If yes, this will invalidate my other comment

Hm, interesting idea. Off hand I can't think of a good way to express that, but I am also no `bash` expert. I can look around some if there is some kind of prior art.

travis-ci/vmtest/helpers.sh
mykola-lysenko: > 

my concern is that if you pass any garbage as a second argument to the function right now it will always behave as closing the fold, third argument does not do anything in this case as well. it is not related to this change, hence I proposed to consider it as a separate PR unrelated to this

danielocfb:
> > Another possible improvement, not sure if this is possible in bash, but can we wrap lists of bash commands with one "foldable" function call? If yes, this will invalidate my other comment
> 
> Hm, interesting idea. Off hand I can't think of a good way to express that, but I am also no `bash` expert. I can look around some if there is some kind of prior art.

I looked around a bit, but couldn't exactly find any suitable syntactical constructs. If we feel strongly about this, the best way forward may be to switch to using Python instead of `bash`. I suppose then we could use `with` blocks.

PR #97, opened 2022-07-21
Remove "temporary" patches

danielocfb:
Both patches were meant to be temporary and, by now, have landed
upstream. The patch application is a no-op in all recent CI runs because
of that. Remove them.


danielocfb:
Here is an example run that shows that patches are just no-ops: https://github.com/kernel-patches/bpf/runs/7450670576?check_suite_focus=true (can't link to the operation in question directly, though).

PR #100, opened 2022-07-26
Consolidate all workflow operations into test.yml

danielocfb:
In my opinion, it is confusing to have half the operations being
performed in CI defined in the "actual" workflow file and the other half
in a separate action. Neither is the action reused by anything nor is
the functional split necessarily intuitive. GitHub Action's enforced
directory structure isn't exactly helping the matter either.
Let's just consolidate everything into `workflows/test.yml`. As a pleasant
side-effect, this also makes the steps being performed in individual CI
runs much more obvious, because we effectively remove one layer of
"folding".


danielocfb:
Still waiting for test run [here](https://github.com/kernel-patches/bpf/actions/runs/2741467069) to finish.

mykola-lysenko (approved):

PR #101, opened 2022-07-27
Use deny/allow lists from upstream

danielocfb:
So far we have relied on allow/deny lists maintained in this repository
to decide which tests to explicitly include/exclude from running in CI.
With recent changes [0] this information is now available in upstream
Linux.
As such, this change switches us over to using the upstream
configuration and removes the locally maintained one.

[0] https://lore.kernel.org/bpf/165893461358.29339.11641967418379627671.git-patchwork-notify@kernel.org/T/#m2a97b0ea9ef0ddee7a53bbf7919e3f324b233937


mykola-lysenko:
Change looks good, thanks Daniel!

I restarted ubuntu-latest with gcc build because previous one was plagued with RCU issues (stashed in https://github.com/kernel-patches/vmtest/runs/7549843468?check_suite_focus=true).

Once tests pass, will approve the change

mykola-lysenko (approved):

danielocfb:
Merging, given that all tests have passed now. Thanks for the review!

PR #102, opened 2022-07-28
Revert "Use deny/allow lists from upstream"

danielocfb:
We are seeing a bunch of timeouts with LLVM toolchain in CI ([example](https://github.com/kernel-patches/bpf/actions/runs/2754779691)). Given the timing (some test runs are 3h old), the only candidate for the breakage I can think of is: https://github.com/kernel-patches/vmtest/pull/101

Let's back out the change and then see what's going on.


danielocfb:
Let's roll with https://github.com/kernel-patches/vmtest/pull/104 instead.

PR #103, opened 2022-07-28
Run on self-hosted runners

danielocfb:
We have seen cases of timeouts in the CI where a GitHub-hosted runner
was used for test execution and that slowed down everything from setup
to kernel build to running of tests.
Given that we have sufficient EC2 runners, which are beefier than
GitHub-hosted ones and, hence, can execute runs in a more timely fashion
and with fewer chances for timeouts, let's make sure to only use
self-hosted runners.
The self-hosted label is automatically assigned to all self-hosted
runners [0]. With this change we make sure to execute jobs only where
this label is present.

[0] https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#choosing-self-hosted-runners

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,15 +17,15 @@ jobs:
       matrix:
         include:
           - kernel: 'LATEST'
-            runs_on: ubuntu-latest
+            runs_on: [ubuntu-latest, self-hosted]
             arch: 'x86_64'
             toolchain: 'gcc'
           - kernel: 'LATEST'
-            runs_on: ubuntu-latest
+            runs_on: [ubuntu-latest, self-hosted]
             arch: 'x86_64'
             toolchain: 'llvm-15'
           - kernel: 'LATEST'
-            runs_on: z15
+            runs_on: [z15, self-hosted]
             arch: 's390x'
             toolchain: 'gcc'
     env:


danielocfb:
The CI failures is spurious and not caused by this change. I am still investigating as it was seen in more than one run. We can retry, but it seems like a waste of time in this specific case, because the change is syntactically checked by GitHub when running the actions and has nothing to do with the actual tests. So I do think that we should move ahead and merge this change to mitigate any potential timeout issues from pull request "storms".

mykola-lysenko (approved):

qmonnet:
Hi @danielocfb,

This change makes it harder to run the CI on other GitHub accounts that do not have self-hosted runners around (use case: I'm occasionally testing my changes with the CI by picking up the file from this repo, see e.g. https://github.com/qmonnet/linux/pull/14). I can add a commit to my PR to revert this change, but I was wondering if there would be a simpler way? Any chance the `self-hosted` label could be added conditionally, e.g. if the repo is under `kernel-patches/` maybe?

danielocfb:
Hi @qmonnet,

> This change makes it harder to run the CI on other GitHub accounts that do not have self-hosted runners around (use case: I'm occasionally testing my changes with the CI by picking up the file from this repo, see e.g. [qmonnet/linux#14](https://github.com/qmonnet/linux/pull/14)). I can add a commit to my PR to revert this change, but I was wondering if there would be a simpler way? Any chance the `self-hosted` label could be added conditionally, e.g. if the repo is under `kernel-patches/` maybe?

You are right, that's one of the side-effects. The reason I concluded that this is acceptable is because:
- testing on s390 machines never worked for a similar reason: the workers are self-hosted and GitHub doesn't have anything eligible
- GitHub workers frequently hit timeouts and so were not exactly reliable for the workflows in question

Basically, the experience was bad and it was never a proper supported workflow.

I think all you'd have to do to get a full run during testing is to create an actual pull request against this repository. Nobody is monitoring pull requests here so I don't believe that spam is an issue. I think a draft pull request would even suffice in case you are worried that somebody would accidentally merge. You can also just keep a pull request open during development and force push to the backing branch. It's certainly what I am doing right now. Have you tried that?

We can certainly check whether conditional tags are supported, but because the experience was so sub-par, I am not sure we really want to go down that road -- at least assuming that the suggested approach works for you.

Let me know what you think.

qmonnet:
> You are right, that's one of the side-effects. The reason I concluded that this is acceptable is because:
> 
>  * testing on s390 machines never worked for a similar reason: the workers are self-hosted and GitHub doesn't have anything eligible

Right, s390 won't work on GitHub in any case

>  * GitHub workers frequently hit timeouts and so were not exactly reliable for the workflows in question

Is this something that can be addressed? I don't know if that timeout can be configured? Asking out of curiosity.

> Basically, the experience was bad and it was never a proper supported workflow.

OK, thanks a lot for the context. I can understand that.

> I think all you'd have to do to get a full run during testing is to create an actual pull request against this repository. Nobody is monitoring pull requests here so I don't believe that spam is an issue. I think a draft pull request would even suffice in case you are worried that somebody would accidentally merge. You can also just keep a pull request open during development and force push to the backing branch. It's certainly what I am doing right now. Have you tried that?

A bit of context of my side as well: I've run the CI on my own Linux fork a few times to test patchsets. It's nowhere close to the amount of runs we do for the patches from the mailing list. But since I did it a few times, I had made a script to automate the creation of the PR, and I was considering [blogging about that](https://github.com/qmonnet/whirl-offload/blob/wip/bpf-ci-check/_posts/2022-08-05-bpf-ci-check.md) - how to run the CI on one's fork and how to use that script.

So for my own usage, I don't really mind either reverting the change from this PR locally, or opening a PR against kernel-patches/vmtest and update it each time I need to run checks. But if I publish my post, I don't feel comfortable asking people to open PRs against the current repo. I'm pretty sure Andrii at least sees the new PRs, because I've seen him answer promptly to a few of them in the past. It would also mean more people using the runners that you finance (although that's more a matter of principle, I don't really expect to convert a lot of readers, if at all, to running their patches on the CI here).

On a side note, we have also mentioned internally the possibility of running the CI on cilium/linux (having kbuildbot running against the repo already), but this would possibly require some other adjustments (we haven't really looked into it yet), I don't think the change for the self-hosted runners is blocking but I just wanted to illustrate there could be more users in the future - assuming you don't mind other projects reusing the work from this repo.

> We can certainly check whether conditional tags are supported, but because the experience was so sub-par, I am not sure we really want to go down that road -- at least assuming that the suggested approach works for you.
> 
> Let me know what you think.

I don't know, it depends on how bad the experience is with GitHub runners I guess. I've run the CI on my fork a few times without too much trouble, so I figured it was mostly OK, but you know it better than I do. Ideally a solution with “self-hosted by default” and an option to overwrite this would sound nice, but we can't really pass options from the command line here, so I guess this would also come down to adding another commit to edit a file and pass that option. Otherwise, adding `self-hosted` only when the workflow runs on kernel-patches would make sense to me, given that this is the condition to have access to the self-hosted runners (as far as I understand?), but would probably require some documentation or at least some comment in the YAML to warn users about the possible timeouts?

danielocfb:
> Is this something that can be addressed? I don't know if that timeout can be configured? Asking out of curiosity.

Yes, the timeout is adjustable. We should be able to bump it up.

> > I think all you'd have to do to get a full run during testing is to create an actual pull request against this repository. Nobody is monitoring pull requests here so I don't believe that spam is an issue. I think a draft pull request would even suffice in case you are worried that somebody would accidentally merge. You can also just keep a pull request open during development and force push to the backing branch. It's certainly what I am doing right now. Have you tried that?
> 
> A bit of context of my side as well: I've run the CI on my own Linux fork a few times to test patchsets. It's nowhere close to the amount of runs we do for the patches from the mailing list. But since I did it a few times, I had made a script to automate the creation of the PR, and I was considering [blogging about that](https://github.com/qmonnet/whirl-offload/blob/wip/bpf-ci-check/_posts/2022-08-05-bpf-ci-check.md) - how to run the CI on one's fork and how to use that script.
> 
> So for my own usage, I don't really mind either reverting the change from this PR locally, or opening a PR against kernel-patches/vmtest and update it each time I need to run checks. But if I publish my post, I don't feel comfortable asking people to open PRs against the current repo. I'm pretty sure Andrii at least sees the new PRs, because I've seen him answer promptly to a few of them in the past. It would also mean more people using the runners that you finance (although that's more a matter of principle, I don't really expect to convert a lot of readers, if at all, to running their patches on the CI here).

Others (@anakryiko @mykola-lysenko) should weigh in as well, but my take is that the CI is actively being worked on. As such, changes are to be expected and it's possible that your script breaks or workflows you described become dysfunctional.

We've certainly mentioned the possibility of giving users easy access to running CI themselves (and personally I'd love to see that feature), but it's not yet clear whether that will ultimately happen (last I checked that wasn't an explicitly stated goal) or how and where it will run.

> On a side note, we have also mentioned internally the possibility of running the CI on cilium/linux (having kbuildbot running against the repo already), but this would possibly require some other adjustments (we haven't really looked into it yet), I don't think the change for the self-hosted runners is blocking but I just wanted to illustrate there could be more users in the future - assuming you don't mind other projects reusing the work from this repo.

For such a use case I'd intuitively say we should rather make the scripts and CI code available and ready for re-use as opposed to hooking into the entire infrastructure we currently have. Then you could self-host and decide what kind of runners to use. In its current incarnation I am not sure there is enough flexibility for either, though. Basically nothing has been designed with reuse in one form or another in mind from what I can tell.
That being said, we are also looking into onboarding more sub-systems to the CI, but haven't finalized the details. Most likely a lot of that would run under the umbrella of Meta and as such there is little or no conflict with using existing runners, even if self-hosted.

danielocfb:
Let me get a few other changes ironed out and then I will look into adding back the tag. I am pretty certain it can be done conditionally. Will have to see about the timeout. I think the same logic can be applied there.

qmonnet:
@danielocfb Apologies for the delay of the answer.

> We've certainly mentioned the possibility of giving users easy access to running CI themselves (and personally I'd love to see that feature), but it's not yet clear whether that will ultimately happen (last I checked that wasn't an explicitly stated goal) or how and where it will run.

Alright, I do keep in mind that things are moving, and that the current CI is not a stable interface ready to be reused on large projects without maintenance and adjustments.

> For such a use case I'd intuitively say we should rather make the scripts and CI code available and ready for re-use as opposed to hooking into the entire infrastructure we currently have. Then you could self-host and decide what kind of runners to use. In its current incarnation I am not sure there is enough flexibility for either, though. Basically nothing has been designed with reuse in one form or another in mind from what I can tell. That being said, we are also looking into onboarding more sub-systems to the CI, but haven't finalized the details. Most likely a lot of that would run under the umbrella of Meta and as such there is little or no conflict with using existing runners, even if self-hosted.

It would be great to have all these elements directly available for reuse. I suppose we'll come to that in time, especially if more subsystems start using the CI indeed.

On a related note, I wanted to let you know that I submitted (and got accepted) a session on the topic at the next Linux Plumbers conference ([link](https://lpc.events/event/16/contributions/1369/)), I don't know if you'll be there. The idea is to give an overview of the BPF CI, of how I've been reusing it so far, and of how other projects might reuse it in the future. Ideally I'd like to trigger some discussion around that, although I certainly don't mean to claim credit for the work you (as a team) have been doing here, or to dictate the future steps for the CI, or even to tell everyone to rush and hook their projects onto the CI. I haven't done the slides yet, but if you wished to have a look at them beforehand or to discuss this more, this is totally doable.

danielocfb:
> On a related note, I wanted to let you know that I submitted (and got accepted) a session on the topic at the next Linux Plumbers conference ([link](https://lpc.events/event/16/contributions/1369/)), I don't know if you'll be there. The idea is to give an overview of the BPF CI, of how I've been reusing it so far, and of how other projects might reuse it in the future. Ideally I'd like to trigger some discussion around that, although I certainly don't mean to claim credit for the work you (as a team) have been doing here, or to dictate the future steps for the CI, or even to tell everyone to rush and hook their projects onto the CI. I haven't done the slides yet, but if you wished to have a look at them beforehand or to discuss this more, this is totally doable.

That is great to hear! Yes, I will be at Linux Plumbers. Will definitely try to attend the session. I'd also be happy to take a look at the slides once you have something.

PR #104, opened 2022-07-29
Add support for allow/deny list additions

danielocfb:
We would like to have the possibility of quickly adding additional allow
or deny lists compared to what the upstream kernel already provides.
This change introduces the infrastructure for doing so. It also brings
back the non-architecture specific deny list removed by fba2b3da955c8
("Use deny/allow lists from upstream") which constitutes the difference
over what upstream Linux now contains, as we have seen increased
flakiness in recent CI runs.

### DIFF
--- a/travis-ci/vmtest/configs/DENYLIST
+++ b/travis-ci/vmtest/configs/DENYLIST
@@ -1,0 +1,5 @@
+btf_dump/btf_dump: syntax
+kprobe_multi_test/bench_attach
+core_reloc/enum64val
+core_reloc/size___diff_sz
+core_reloc/type_based___diff_sz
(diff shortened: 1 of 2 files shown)


danielocfb:
https://github.com/kernel-patches/bpf/runs/7573473629?check_suite_focus=true is a test run with the change and where I included logging of the deny list used. We can see that the newly denied tests are correctly contained in it:
```
DENYLIST: ,get_stack_raw_tp,stacktrace_build_id_nmi,stacktrace_build_id,task_fd_query_rawtp,varlen,btf_dump/btf_dump: syntax,kprobe_multi_test/bench_attach,core_reloc/enum64val,core_reloc/size___diff_sz,core_reloc/type_based___diff_sz,
```

mykola-lysenko (approved):

PR #105, opened 2022-07-29
ci: fix printing of currently used configuration

mykola-lysenko:
We started to use array of tags during the CI
configuration print, but GH actions do not expand
array and just print 'Array' literal which is
confusing


danielocfb (approved):
Thanks for the fix! The one test failure should hopefully be addressed by us switching back to using not-quite-minimized kernel configuration (https://github.com/libbpf/ci/pull/23), but it's certainly not caused by this change.

PR #106, opened 2022-07-29
Topic/revert upstream deny allow pr

danielocfb:
CI runs on the `bpf` (as opposed to `bpf-next`) tree fail with recent changes, because there we do not (yet) have the upstream configuration present. That is mitigated by https://github.com/libbpf/ci/pull/23, but conceptually the same issue exists for deny/allow lists. The difference there is that we won't fail when a file in question does not exist (because it may legitimately not exist). However, as a result we may end up running tests that are flaky.
Let's back out the change switching everything over to using the upstream deny/allow lists (and the one depending on it) to be sure that we will work with the correct lists and not have spurious test failures.

### DIFF
--- a/travis-ci/vmtest/configs/DENYLIST
+++ b/travis-ci/vmtest/configs/DENYLIST
@@ -1,5 +1,0 @@
-btf_dump/btf_dump: syntax
-kprobe_multi_test/bench_attach
-core_reloc/enum64val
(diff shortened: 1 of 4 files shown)


danielocfb:
Uhm...I am not sure why the s390x step was "skipped" in the middle of the run.

mykola-lysenko (approved):

PR #110, opened 2022-08-04
Rename config-latest.* to config.*

danielocfb:
The repository no longer stores configurations for past kernel versions.
As such, the "-latest" suffix is superfluous. It's also no longer what
we expect once https://github.com/libbpf/ci/pull/25 lands.
Remove it by renaming the files accordingly.


danielocfb:
CI runs could be a bit tricky, because the changes depend on each other in a circular fashion... :-(

danielocfb:
Changed approach for https://github.com/libbpf/ci/pull/25 to honor both `config` and `config-latest`. Will later remove support for `config-latest`.

danielocfb:
The change is trivial and all tests have passed. Going ahead with the merge.

PR #119, opened 2022-08-17
Rename travis-ci/ directory to ci/

danielocfb:
We are no longer using Travis. As such, we should move away from a lot
of CI functionality located in a folder called travis-ci/. This change
renames the travis-ci/ directory to the more generic ci/.
To preserve backwards compatibility until all "consumers" have
transitioned, we add a symbolic link called travis-ci back. It will be
removed in the near term future.


danielocfb:
I checked KPD and have not found any references to the `travis-ci` directory (anymore).

mykola-lysenko (approved):

danielocfb:
Hm, it seems as if CI is choking on this change. E.g., https://github.com/kernel-patches/bpf/actions/runs/2883830082
Will revert it and see what's going on.

danielocfb:
So I can't really say that I understand the issue we hit fully. But after a bit of investigatory work and in the hope to reproduce, I managed to piece the following together:
- looking at https://github.com/kernel-patches/bpf/runs/7902852711?check_suite_focus=true we can see that we checked out 6884ed6fe662f2afb7a12a0a1ba3c7b33bf83a3d (somewhere burried in the logs)
- if I do `git fetch https://github.com/kernel-patches/bpf.git 6884ed6fe662f2afb7a12a0a1ba3c7b33bf83a3d` and then checkout 6884ed6fe662f2afb7a12a0a1ba3c7b33bf83a3d, we can see that there is no `travis-ci` symlink present
- if I look at https://github.com/kernel-patches/vmtest/commit/cd79e2cdb1dcb8e52fc8e18bfe28b4ceb395a0d6 we can see that the symbolic link exists
- I can't reproduce the issue with any manual runs I've scheduled
- so that really only leaves KPD as the source of problem

PR #123, opened 2022-08-18
Add patch adding lru_bug to deny list

danielocfb:
Recently bpf got merged into bpf-next. That merge pulled in change
de7b9927105b ("selftests/bpf: Add test for prealloc_lru_pop bug") but
did not include 27e23836ce22 ("selftests/bpf: Add lru_bug to s390x deny
list") yet.
As a result, we are now seeing the lru_bug selftest run (and fail) on CI
runs of bpf-next.
Backport the fix to get CI runs back into green state.


anakryiko (approved):

danielocfb:
The patch got applied successfully. `lru_bug` is shown in the denylist used. The remaining test failure is unrelated. Merging.

danielocfb:
So now we changed upstream `bpf-next` again to add other tests to the deny list and that means the patch no longer applies. What a royal mess. Guess we'll have to add it to the local list then...

PR #125, opened 2022-08-19
Temporarily pin libbpf/ci revision to use

danielocfb:
We want to rename the travis-ci/ directory to the more generic ci/, as
we are no longer using Travis. When we earlier landed that change [0] we
saw issues because the libbpf/ci repository, which contains the majority
of the actions we use, still expects contents in travis-ci/.
To make the change atomic, this change temporarily pins the revision of
libbpf/ci that we use. Once libbpf/ci has been adjusted to work with the
ci/, we can switch this repository over atomically.

[0] https://github.com/kernel-patches/vmtest/pull/119

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,7 @@ jobs:
       llvm: ${{ steps.llvm-toolchain-impl.outputs.version }}
     steps:
       - id: llvm-version
-        uses: libbpf/ci/get-llvm-version@master
+        uses: libbpf/ci/get-llvm-version@da44c0b6ee29ec53776c1303f7ac790b7e337db5
       - id: llvm-toolchain-impl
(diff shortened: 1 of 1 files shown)


mykola-lysenko:
why symlinks did not work though?

danielocfb:
> why symlinks did not work though? I thought it was the point to add them

It was. And I am not sure. I suspect it's some intricacy with whether we linked `ci/` or `ci` or the fact that they are second class citizens whose handling has to be special cased half the time. I don't feel strongly about either approach, but I can get to the bottom of it if that's the preferred way.

danielocfb:
Will look into whether there is a different work around. It gets messy if we have changes landing in parallel, such as https://github.com/kernel-patches/vmtest/pull/126. Sigh.

PR #127, opened 2022-08-22
Run selftests as part of a separate job

danielocfb:
This change factors out the running of selftests into a separate job.
Doing so is a precursor to being able to running individual tests in
parallel, which in turn should decrease total CI run time and increase
resiliency to kernel panics encountered in one run.
To make that possible we basically have to have well defined output of
the build step that can be consumed by the test step. At this point,
this only means preserving kernel configuration, the selftests/bpf
folder contents, and the vmlinux image itself, and making them available
to tests. As such, we package them up as artifacts. We manually tar
those in order to preserve permissions [0] but also to potentially curb
any spurious 429 HTTP errors ("too many requests") caused by too many
individual files being uploaded if selftest/bpf folder contents grow
[1].
In its current form, we split the two stages after building the
selftests. An alternative split would be after generation of the Qemu
image. Based on arguably limited testing, the increase in artifact
upload and download size (almost doubling; presumably because of
additional files in the image) negates any speed ups achieved by
performing image generation as part of the build. It's also slightly
less natural, because bpftool tests are actually run as part of image
generation, and so part of the tests would be run in the build stage.

Note that because of GitHub Actions' limitations, scheduling of jobs
will not be optimal. Specifically, *any* test job is only eligible to
run once *all* build jobs have finished. Strictly speaking, the s390x
test job would only have to depend on the s390x build job, for example,
but that cannot be expressed in GitHub Actions -- at least not when
using job matrices. We could potentially fix that by "unfolding" the
matrix manually, but that quickly becomes cumbersome because GitHub
Actions does not support YAML Anchors either [2]. While that shortcoming
will have an impact on total CI run time, it should not have any bearing
on runner utilization (because the moment a job is finished the runner
can move on to serving another one). That increase will get mitigated
somewhat once we run tests in parallel.
Lastly, this patch also indirectly addresses the timeout issues
discussed earlier [3]. Specifically, because both the build and test
steps now have a 100 minute timeout, we are less likely to hit either,
even when executing on a slow runner.

[0] https://github.com/actions/upload-artifact#maintaining-file-permissions-and-case-sensitive-files
[1] https://github.com/actions/upload-artifact#too-many-uploads-resulting-in-429-responses
[2] https://github.com/actions/runner/issues/1182
[3] https://github.com/kernel-patches/vmtest/pull/103

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,10 +3,6 @@ name: bpf-ci
 on:
   pull_request:
 
-concurrency: 
-  group: ci-test-${{ github.head_ref }}
-  cancel-in-progress: true
-
 jobs:
   llvm-toolchain:
     runs-on: ubuntu-latest
@@ -14,7 +10,7 @@ jobs:
       llvm: ${{ steps.llvm-toolchain-impl.outputs.version }}
     steps:
       - id: llvm-version
-        uses: libbpf/ci/get-llvm-version@master
+        uses: danielocfb/libbpf-ci/get-llvm-version@topic/ci
       - id: llvm-toolchain-impl
         shell: bash
         run: echo "::set-output name=version::llvm-${{ steps.llvm-version.outputs.version }}"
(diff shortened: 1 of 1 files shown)


danielocfb:
Just for testing purposes. Please ignore.

PR #130, opened 2022-08-25
Only require self-hosted runners in kernel-patches organization

danielocfb:
The kernel-patches organization hosts their own GitHub Actions runners
which generally are much quicker at running individual CI jobs. To that
end, aecee1ab23bf ("Run on self-hosted runners") prevented the running
on GitHub-hosted ones altogether.
We would still provide the possibility for users of the BPF CI system to
trigger runs without having to create a pull request against a
repository in said org [0] (which, in the future, could conceivably require
approval before executing actions).
To enable this use case, this change requires the self-hosted tag for
build and selftest jobs only when running inside the kernel-patches
organization.
Furthermore, because GitHub does not have any s390x runners, do not test
on these when outside of kernel-patches organization.

[0] https://github.com/kernel-patches/vmtest/pull/103

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,6 +35,16 @@ jobs:
             {"kernel": "LATEST", "runs_on": ["ubuntu-latest", "self-hosted"], "arch": "x86_64", "toolchain": "${{ needs.llvm-toolchain.outputs.llvm }}"},
             {"kernel": "LATEST", "runs_on": ["z15", "self-hosted"], "arch": "s390x", "toolchain": "gcc"},
           ]
+
+          if "${{ github.repository_owner }}" != "kernel-patches":
+            # Outside of kernel-patches, remove the self-hosted label and skip
+            # any testing on s390x, as no suitable runners will be available.
+            for idx in range(len(matrix) - 1, -1, -1):
+              if "z15" in matrix[idx]["runs_on"]:
+                del matrix[idx]
+              else:
+                matrix[idx]["runs_on"].remove("self-hosted")
+
           build_matrix = {"include": matrix}
           print(f"::set-output name=build_matrix::{dumps(build_matrix)}")
 


danielocfb:
I scheduled the following runs:
- https://github.com/kernel-patches/bpf/actions/runs/2929563672 (inside the `kernel-patches` organization)
- https://github.com/danielocfb/kernel-patches-bpf/actions/runs/2929552201 (outside the organization)

danielocfb:
cc @qmonnet

mykola-lysenko:
Thanks Daniel!

Does it mean that if this repo / kernel_patches/bpf are forked and test run is scheduled it will only run on GH runners without any way to test ibm code?

danielocfb:
> Thanks Daniel!
> 
> Does it mean that if this repo / kernel_patches/bpf are forked and test run is scheduled it will only run on GH runners without any way to test ibm code?

Yes, unless you make an explicit pull request against `kernel-patches/bpf`.

> But all the pull requests against this and kernel_patches/bpf repos are still going to be tested with z15?

Correct.

mykola-lysenko (approved):

.github/workflows/test.yml
```diff
@@ -31,10 +31,23 @@ jobs:
...
           matrix = [
-            {"kernel": "LATEST", "runs_on": ["ubuntu-latest", "self-hosted"], "arch": "x86_64", "toolchain": "gcc"},
-            {"kernel": "LATEST", "runs_on": ["ubuntu-latest", "self-hosted"], "arch": "x86_64", "toolchain": "${{ needs.llvm-toolchain.outputs.llvm }}"},
-            {"kernel": "LATEST", "runs_on": ["z15", "self-hosted"], "arch": "s390x", "toolchain": "gcc"},
+            {"kernel": "LATEST", "runs_on": ["ubuntu-latest"], "arch": "x86_64", "toolchain": "gcc"},
+            {"kernel": "LATEST", "runs_on": ["ubuntu-latest"], "arch": "x86_64", "toolchain": "${{ needs.llvm-toolchain.outputs.llvm }}"},
+            {"kernel": "LATEST", "runs_on": ["z15"], "arch": "s390x", "toolchain": "gcc"},
           ]
+
+          if "${{ github.repository_owner }}" == "kernel-patches":
```
mykola-lysenko, line 39: one thing I thought about after approving.

what do you think about keeping self-hosted in the array and then have one loop filtering for z15 / self-hosted and removing them.

it will simplify the logic a bit

danielocfb: Sure, we can do that.

qmonnet (approved):
This is great, thanks a lot for adjusting this! Much appreciated

PR #131, opened 2022-08-29
Explicitly provide kernel version to use

danielocfb:
We currently implicitly rely on the default value for the kernel version
of the prepare-rootfs action. That's counter intuitive, because we
explicitly carry around the kernel version to use in our job matrix.
Let's make the specification explicit instead.


danielocfb:
It's a trivial change with full test coverage. All tests have passed. Merging in order to move forward.

PR #135, opened 2022-09-23
ci: enable parallel test_progs execution

mykola-lysenko:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,7 +53,7 @@ jobs:
           build_matrix = {"include": matrix}
           print(f"::set-output name=build_matrix::{dumps(build_matrix)}")
 
-          tests = ["test_progs", "test_progs_no_alu32", "test_maps", "test_verifier"]
+          tests = ["test_progs", "test_progs_parallel", "test_progs_no_alu32", "test_progs_no_alu32_parallel", "test_maps", "test_verifier"]
           test_matrix = {"include": [{**config, **{"test": test}}
                                       for config in matrix
                                         for test in tests]}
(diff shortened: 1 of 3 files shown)


mykola-lysenko:
After retries, all of the checks are passed.

We can add Song's fix temporarily as I observed quite a few lockdep failures in the places related to his fix.

mykola-lysenko:
added Song's patch as a separate commit, let's see the test result

danielocfb (commented):
Looks good for the most part, thanks! Personally, I'd have introduced an intermediate commit that factors out `test_progs_helper` and then a second one doing the rest.

The only real issue I have is that I was under the impression we wanted to have the parallel tests configured in such a way that if they fail the entire run won't fail (until we have gotten more confidence in parallel execution). It's hard for me to tell what the current behavior is, because not just parallel tests failed in the run, but I think we may need [`jobs.<job_id>.steps[*].continue-on-error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) to achieve said behavior.

Please let me know what you think.

danielocfb:
> but I think we may need [`jobs.<job_id>.steps[*].continue-on-error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) to achieve said behavior

Actually, I am no longer sure it works the way we expected it to, see https://github.com/actions/toolkit/issues/399#issuecomment-674618750. Worth trying it out.

danielocfb (changes requested):
In addition to the comment I made (which we probably should clarify), the change can't be merged currently because of merge conflicts. So I am putting this back into your queue, @mykola-lysenko .

danielocfb:
Let me close this pull request and create a new one.

PR #136, opened 2022-10-05
Rename travis-ci/ directory to ci/

danielocfb:
We are no longer using Travis. As such, we should move away from a lot of CI functionality located in a folder called travis-ci/. This change renames the `travis-ci/` directory to the more generic `ci/`. To preserve backwards compatibility until all "consumers" have transitioned, we add a symbolic link called travis-ci back. It will be removed in the near term future.

This change is a respin of cd79e2cdb1dc ("Rename travis-ci/ directory to ci/"), which got backed out because of issues that turned out to reside in Meta's internal Kernel Patches Daemon. These issues have since been fixed and an updated version deployed.


danielocfb:
Note to Meta folks: I did test with the new version of KPD in a devvm and saw correct update of `travis-ci` from directory to symbolic link.

PR #137, opened 2022-10-06
[artifacts] Use zstd to compress artifacts

chantra:
zstd offers better compression ration and is multi-threaded so it can leverage the CPUs available in the host to compress/decompress faster. With better compression ratio, the size of the artifacts  uploaded/downloaded will be smaller hence faster upload/download.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -117,8 +117,8 @@ jobs:
               tools/include/ \
               tools/bpf/bpftool/";
           fi
-
-          tar -czf vmlinux-${{ matrix.arch }}-${{ matrix.toolchain }}.tar.gz \
+          # zstd is installed by default in the runner images.
+          tar -cf - \
             .config \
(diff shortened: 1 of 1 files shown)


chantra:
this is dependent on all z15 runners being running the docker containers generated by https://github.com/libbpf/libbpf/pull/595

I have manually pushed this change to the kernel-patches/vmtest z15 runners to get CI to work.

chantra:
the one that fails seems totally unrelated to thisa:
https://github.com/kernel-patches/vmtest/actions/runs/3199125414/jobs/5224918722#step:6:13341
> watchdog: BUG: soft lockup - CPU#6 stuck for 2630s! [test_progs-no_a:160]


https://github.com/kernel-patches/vmtest/actions/runs/3199125414/jobs/5224918298#step:6:5921
>   [  370.598931] #PF: supervisor instruction fetch in kernel mode

PR #139, opened 2022-10-07
ci: Make run_selftests.sh read test names from boot parameters

chantra:
When running the tests with qemu, we used to bake the name of the tests to run as part of the root image.
this forces us to have each individual test_suite run build the image (currently 4 times).
All of them are still copying all tests in the image though.

In order to avoid this duplication of work, we can build the image once with all needed binaries, and then let the step running the tests pass it as an argument.
The way to do that from outside the VM, is to leverage the boot parameters and have some logic in the run_selftests.sh script to extract those tests from /proc/cmdline.

Having a root image where we can instruct what test to run will also allow us to essentially run the test on any platform able to emulate the target architecture. For instance, we could run aarch64 tests on x86_64 host.

Another advantage is github aciton will pass this image as an artifact from one step to another. This also means that the root image will be readily available for download and we could easily reply the test on the device of our choice.

The next step will be to change libbpf/ci actions to stop adding test in the image and pass them when calling qemu.

Testing the parsing of run_tests from cmdline:
https://gist.github.com/chantra/5c7dd823e2b7c9419ef55d8139e82010

Testing sorting of 2 arrays:
https://gist.github.com/chantra/3c06ff7722568c928b79e3ba5b8ae5a7


chantra:
This is failing because /dev/fd is not symlinked to /proc/self/fd, which break process substitution. I am a diff for `libbpf/libbpf/ci/rootfs/mkrootfs_tweak.sh` to fix that.

chantra:
putting this back in my plate. It seems tests are being run twice.

ci/vmtest/run_selftests.sh
```diff
@@ -81,13 +102,22 @@ echo "ALLOWLIST: ${ALLOWLIST}"
 
 cd ${PROJECT_NAME}/selftests/bpf
 
-if [ $# -eq 0 ]; then
+# populate TEST_NAMES_FROM_BOOT
```
danielocfb, line 105: Not a fan of having a function populate a global variable as a side effect, but I suppose that's just a `bash`-ism we have to live with?

ci/vmtest/run_selftests.sh
```diff
@@ -81,13 +102,22 @@ echo "ALLOWLIST: ${ALLOWLIST}"
...
+read_tests_from_boot_parameters
+# Sort and only keep unique test names from both boot params and arguments
+# test_names will contain a sorted list of uniq tests to be ran.
+readarray -t test_names < \
+    <(printf '%s\0' "${TEST_NAMES_FROM_BOOT[@]}" "$@" | sort -z -u | xargs -0n1)
```
danielocfb, line 110: Would it make more sense to move this logic into `read_tests_from_boot_parameters`? Having duplicated tests doesn't really make sense to me, to begin with.

Can we use long arguments, please? I think code should be read- and understandable without having two `man` pages handy at all times, to the degree that is possible with shell scripts.

ci/vmtest/run_selftests.sh
chantra: yeah, bash does not have many options in that regard. The alternatives are not necessarily less complex.
https://www.linuxjournal.com/content/return-values-bash-functions has some options for returning values, but it does not seem easier to maintain.

ci/vmtest/run_selftests.sh
chantra: > Would it make more sense to move this logic

We can, as long as we make sure to call this last, we should be fine.

> Having duplicated tests doesn't really make sense to me, to begin with.

Agreed, because we have some of the logic that leaves in libbpf/ci (prepare-rootfs/run-qemu) and here, there is a time window when a mix of both passing through a parameter or boot param may happen.
I rather be conservative here and deal with potential dups making it in rather than ending up running the test twice.

> Can we use long arguments, please?

yes, that's doeable for commands that supports long args (e.g sort and xargs in this case). Will update.

danielocfb (approved):

PR #140, opened 2022-10-11
Add deny_namespace to DENYLIST.s390x configuration

Decave:
The deny_namespace selftest was recently added to validate the userns_create LSM hooks. The test does not work on s390x because LSM uses BPF trampoline.
8206e4e95230 ("selftests/bpf: Add selftest deny_namespace to s390x deny list") was added to the bpf-next tree, but the bpf CI runs are still failing because the commit is not in that tree. So as to avoid merge conflicts, let's just blacklist it on vmtest.

Signed-off-by: David Vernet <void@manifault.com>


Decave:
Updated to apply @danielocfb's feedback on https://github.com/libbpf/libbpf/pull/599 to apply to this commit

PR #141, opened 2022-10-11
update actions/checkout to v3

chantra:
Actions runs started showing the error:

> build for x86_64 with llvm-16
> Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/checkout

This change sets actions/checkout to v3, which is an upgrade from v2 in out first use case, and a pinning to v3 vs main for the second use case.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -71,7 +71,7 @@ jobs:
       REPO_ROOT: ${{ github.workspace }}
       REPO_PATH: ""
     steps:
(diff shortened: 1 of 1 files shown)


danielocfb (approved):

danielocfb:
Failure is unrelated. Merging.

PR #143, opened 2022-10-12
Deprecated set output

chantra:


.github/workflows/test.yml
```diff
@@ -29,6 +29,12 @@ jobs:
         shell: python3 -I {0}
         run: |
           from json import dumps
+          import os
+
+          # write an environemt variable in the GH env file.
+          def setOutput(name, value):
+            with open(os.getenv("GITHUB_OUTPUT"), "a") as f:
```
danielocfb, line 36: Let's maybe use a doc comment? Also, can we follow [PEP 8](https://peps.python.org/pep-0008/#function-and-variable-names) with respect to naming functions?

```
          def set_output(name, value):
            """Write an output variable to the GitHub output file."""
            with open(os.getenv("GITHUB_OUTPUT"), "a") as f:
```

Nit: Also not sure if referring to an "environment" variable is the correct terminology.

danielocfb (approved):
Overall looks good, but please consider addressing my comments before merging.

.github/workflows/test.yml
chantra: > Nit: Also not sure if referring to an "environment" variable is the correct terminology.

Yeah, I got originally mislead by their link in https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/ that lead
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files

searching for "set-output" was leading to something totally different and lead me to think they recommended putting everything in env variables.

PR #150, opened 2022-10-25
Topic/parallel

danielocfb:


ci/vmtest/run_selftests.sh
```diff
@@ -54,21 +48,34 @@ read_test_names() {
...
 }
 
-test_progs() {
-  foldable start test_progs "Testing test_progs"
+test_progs_helper() {
+  local suffix="$1"
+  local args="$2"
+  local selftest="test_progs${suffix}"
+
+  foldable start test_progs${suffix} "Testing test_progs${suffix}"
```
mykola-lysenko, line 56: minor: can we reuse $selftest variable here?

ci/vmtest/run_selftests.sh
```diff
@@ -54,21 +48,34 @@ read_test_names() {
     foldable end read_test_names
 }
 
-test_progs() {
-  foldable start test_progs "Testing test_progs"
+test_progs_helper() {
+  local suffix="$1"
```
mykola-lysenko, line 52: minor: if we address below comment, do we need suffix variable or can move it there?

mykola-lysenko:
thanks Daniel, other then two minor comments, everything looks good. waiting until tomorrow to see tests pass and confirm that failed parallel tests not to fail the overall test result

ci/vmtest/run_selftests.sh
danielocfb: Indeed, we should.

ci/vmtest/run_selftests.sh
danielocfb: Agreed.

mykola-lysenko (approved):

danielocfb:
So the problem I see is parallel runs seem to trigger hangs/deadlocks/livelocks in the kernel causing us to time out. And those timeouts actually manifest as an error after all.

danielocfb:
> So the problem I see is parallel runs seem to trigger hangs/deadlocks/livelocks in the kernel causing us to time out. And those timeouts actually manifest as an error after all.

In addition to the change that enabled panics on soft & hard lockups, I've added a patch that enables hung task detection. From my perspective, that resolves these issues.

@mykola-lysenko can I ask for another review, please?

mykola-lysenko:
> > So the problem I see is parallel runs seem to trigger hangs/deadlocks/livelocks in the kernel causing us to time out. And those timeouts actually manifest as an error after all.
> 
> In addition to the change that enabled panics on soft & hard lockups, I've added a patch that enables hung task detection. From my perspective, that resolves these issues.
> 
> @mykola-lysenko can I ask for another review, please?

so, basically, because we are not going to timeout jobs anymore, they will fail and this will make the run to pass, correct?

mykola-lysenko:
> > > So the problem I see is parallel runs seem to trigger hangs/deadlocks/livelocks in the kernel causing us to time out. And those timeouts actually manifest as an error after all.
> > 
> > 
> > In addition to the change that enabled panics on soft & hard lockups, I've added a patch that enables hung task detection. From my perspective, that resolves these issues.
> > @mykola-lysenko can I ask for another review, please?
> 
> so, basically, because we are not going to timeout jobs anymore, they will fail and this will make the run to pass, correct?

if yes, this sounds good to me

I also see that Alexei accepted your patch to bpf-next, but I guess it is here to cover bpf as well, right?

mykola-lysenko (approved):
Change looks good to me!

Here, we added two diffs as well. One is already upstreamed in bpf-next and will be removed once it is in bpf as well. Another one needs more testing in the CI before upstreaming.

Before merging, let's ensure test run passes and I would love to get my questions answered.

Thanks @d-e-s-o !

danielocfb:
> > ...
> so, basically, because we are not going to timeout jobs anymore, they will fail and this will make the run to pass, correct?

Yes, exactly.

> I also see that Alexei accepted your patch to bpf-next, but I guess it is here to cover bpf as well, right?

That is correct as well. Otherwise the `0002-*` patch would not apply cleanly on `bpf` (and we would not have the soft & hard lockup detection, of course)

PR #152, opened 2022-10-28
Run shellcheck only in vmtest repository

danielocfb:
Our vmtest repository files (including GitHub Actions workflow definitions) get injected into the Linux source code repository in order to trigger CI runs on it. As a result, any workflows defined here will run there as well. That is fine for our actual Linux tests, but not so for the linting of shell scripts: the linter was intended to only cover vmtest shell scripts itself.
Make linting conditional on the repository we are run in to prevent spurious invocations.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -8,6 +8,9 @@ on:
 
 jobs:
   shellcheck:
+    # This workflow gets injected into other Linux repositories, but we don't
+    # want it to run there.
+    if: ${{ github.repository == 'kernel-patches/vmtest' }}
     name: ShellCheck
     runs-on: ubuntu-latest
     steps:


danielocfb:
The shellcheck check still passed. I canceled the "regular" CI run as nothing changed there.

.github/workflows/lint.yml
```diff
@@ -13,7 +13,10 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
-      - name: Run ShellCheck
+      # This workflow gets injected into other Linux repositories, but we don't
```
chantra, line 16: Can you move the condition  just before the actual steps (line 13), I believe this will make the whole job a no-op vs performing some of the steps.

danielocfb: Sure thing, done.

danielocfb:
Let's get this merged to prevent more noise coming from bpf CI then.

PR #154, opened 2022-11-01
Update DENYLIST

chantra:
Add netcnt into ci/vmtest/configs/DENYLIST.x86_64 so it passes on kvm-enabled vm and copy ci/vmtest/configs/DENYLIST.aarch64 from bpf-next.


danielocfb (approved):
Please use a `.diff` for the `bpf-next` commit instead of importing the raw denylist as discussed offline. Other than that the changes look fine to me. Should we open an issue for the `netcnt` investigation?

chantra:
> Should we open an issue for the `netcnt` investigation?

yes, I can create an issueon GH for visibility.

danielocfb:
Is `serial_test_netcnt` having the same issue as `netcnt` and should it be denylisted as well? https://github.com/kernel-patches/bpf/actions/runs/3381075212/jobs/5615096678#step:6:6634

PR #161, opened 2022-11-03
diff: remove .patch files

chantra:
Those are not applied by the tooling, only .diff are.


danielocfb (approved):
Thanks for the cleanup.

PR #162, opened 2022-11-03
workflow: re-arrange runner selection logic

chantra:
If we are not in a vetted repo, only run on x86_64's ubuntu-latest runners (GH runners)
Otherwise, run on runners that have both `self-hosted` and `uname -m` labels.

This will allow to move self-hosted runners at org level and have them being used across the different repos we support under kernel-patches and make better use of the pool of runners.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,9 +56,9 @@ jobs:
             return config
 
           matrix = [
-            {"kernel": "LATEST", "runs_on": ["ubuntu-latest", "self-hosted"], "arch": "x86_64", "toolchain": "gcc"},
-            {"kernel": "LATEST", "runs_on": ["ubuntu-latest", "self-hosted"], "arch": "x86_64", "toolchain": "${{ needs.llvm-toolchain.outputs.llvm }}"},
-            {"kernel": "LATEST", "runs_on": ["z15", "self-hosted"], "arch": "s390x", "toolchain": "gcc"},
+            {"kernel": "LATEST", "runs_on": [], "arch": "x86_64", "toolchain": "gcc"},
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -67,13 +67,17 @@ jobs:
 
           # Only a few repository within "kernel-patches" use self-hosted runners.
           if "${{ github.repository_owner }}" != "kernel-patches" or "${{ github.repository }}" not in self_hosted_repos:
-            # Outside of those repositories, remove the self-hosted label and skip
-            # any testing on s390x, as no suitable runners will be available.
+            # Outside of those repositories, we only run on x86_64 GH hosted runners (ubuntu-latest)
+            # else we run on self-hosted + arch
```
danielocfb, line 71: > else we run on self-hosted + arch

Remove this line? It seems to be what the else branch comments again.

.github/workflows/test.yml
```diff
@@ -56,9 +56,9 @@ jobs:
             return config
 
           matrix = [
-            {"kernel": "LATEST", "runs_on": ["ubuntu-latest", "self-hosted"], "arch": "x86_64", "toolchain": "gcc"},
-            {"kernel": "LATEST", "runs_on": ["ubuntu-latest", "self-hosted"], "arch": "x86_64", "toolchain": "${{ needs.llvm-toolchain.outputs.llvm }}"},
-            {"kernel": "LATEST", "runs_on": ["z15", "self-hosted"], "arch": "s390x", "toolchain": "gcc"},
+            {"kernel": "LATEST", "runs_on": [], "arch": "x86_64", "toolchain": "gcc"},
```
danielocfb, line 59: Can we consider introducing constants for the architectures we support instead of duplicating the strings?

danielocfb (approved):

.github/workflows/test.yml
chantra: slightly over-engineered it with enums.

PR #167, opened 2022-11-08
ci: Enable CI on aarch64

chantra:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,8 +36,9 @@ jobs:
             """
             CPU architecture supported by CI.
             """
-            x86_64 = "x86_64"
+            aarch64 = "aarch64"
             s390x = "s390x"
+            x86_64 = "x86_64"
 
           def set_output(name, value):
             """Write an output variable to the GitHub output file."""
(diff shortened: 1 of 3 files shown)


chantra:
At the time of testing x86_64 with llvm was failing. https://github.com/kernel-patches/vmtest/actions/runs/3423770717 is a run on arm64 only.

danielocfb (approved):
Looks great to me! Big milestone.

Are these additional test failures (new entries in DENYLIST) understood fully? Should we keep the three column format trying to capture the essence of the problem?

chantra:
> Are these additional test failures (new entries in DENYLIST) understood fully?

No, and they should probably be addressed separately.

> Should we keep the three column format trying to capture the essence of the problem?

The 3 column format is not obvious from the file in this repo. Probably the only place where this is made more obvious is https://github.com/kernel-patches/bpf/blob/bpf-next/tools/testing/selftests/bpf/DENYLIST.s390x . Files should probably have a header explaining the format. This can go in a separate diff.

chantra:
I will land this for now.  kernel-patches/vmtest#169  for file format.

PR #170, opened 2022-11-09
Build rootfs image after building kernel

chantra:
This is the complementary part of libbpf/ci#74.

In this change, we build the VM's rootfs right after building the kernel/selftests....

Then, when running qemu, we pass the name of the test we want to run.

The location of root.img was also changed from /tmp to workspace. The reason behind this change is that on bare-metal runners, workspace is mounted on tmpfs and this will help removing some more IOs from those hosts.

libbpf/ci#74 handles the case if root.img is on tmpfs, it will not use the "cache=none" options, which relies on O_DIRECT, and is not available on tmpfs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -259,8 +259,16 @@ jobs:
           toolchain: ${{ matrix.toolchain }}
           kbuild-output: ${{ env.KBUILD_OUTPUT }}
           max-make-jobs: 32
-      - if: ${{ github.event_name != 'push' }}
-        name: Tar artifacts
+      - name: Prepare rootfs
+        uses: libbpf/ci/prepare-rootfs@master
+        with:
+          project-name: 'libbpf'
+          arch: ${{ matrix.arch }}
+          kernel: ${{ matrix.kernel }}
+          kernel-root: '.'
+          kbuild-output: ${{ env.KBUILD_OUTPUT }}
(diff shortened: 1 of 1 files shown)


danielocfb:
Can you share more of the motivation for this change, please? Is it to speed up CI? It seems as if all artifacts are now 30+ MiB larger in size, so intuitively it's not clear whether we gain anything performance-wise, if that's the goal.

I also see that s390 are still failing and you still seemed to be investigating. Should this be converted back to Draft state?

chantra:
> Can you share more of the motivation for this change, please? Is it to speed up CI? It seems as if all artifacts are now 30+ MiB larger in size, so intuitively it's not clear whether we gain anything performance-wise, if that's the goal.
> 

The motivation is 2 folds:
- build the vm once and run it for every tests. We trade bigger artifacts for CPU win. e.g when we use to have 4 test runners building the VM, we now only have the kernel build runner doing this. end to end time should be relatively the same, but we while runners are not creating the vm, they can do other stuff (running test, building kernels).
- second, having the VM directly available from the artifacts to download means that operators can easily access them, download them and run the VM locally on their computer in order to troubleshoot issues.

> I also see that s390 are still failing and you still seemed to be investigating. Should this be converted back to Draft state?

yes

chantra:
With #243 landed, this is not needed anymore.

PR #172, opened 2022-11-11
Use matrix.arch as cache key identifier

danielocfb:
We have seen some architecture related build failures with the incremental build logic in place. E.g.,
https://github.com/kernel-patches/bpf/actions/runs/3441158955/jobs/5740411241#step:10:49

We can see that we picked up the cache from
kbuild-output-Linux-gcc-1668121889-67377655e58ae24ef2125c05e4dcfeb2e59d64dc. That cache was actually created on x86_64, but the job is running aarch64.
The reason that is done is because we are using `runner.os` as part of the key, when really we should be using `matrix.arch`. Let's use the latter instead. We only currently have one OS we test on.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -148,9 +148,10 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ${{ env.KBUILD_OUTPUT }}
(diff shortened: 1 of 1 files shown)


danielocfb:
Going ahead with the merge to hopefully get CI back on track over night. Right now we need to manually clear caches or `aarch64` builds will fail.

PR #173, opened 2022-11-11
Use a different branch key depending on event received

danielocfb:
With commit 0f78398f5f30 ("Include base branch name in cache key identifier") we still have one problem: GitHub uses different attributes to represent the various branches/refs of interest in different contexts. E.g., in pull requests it populates the `base_ref` field, whereas for pushes it is empty. Similarly, `ref_name` refers to different things.
With this change we "generate" the branch name to include in the cache key identifier with keeping that into account. Basically, when we want to create the cache (on the "push" path) we can rely on `ref_name`, whereas when we are interested in pulling the cache (pull request), we need to use `base_ref`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -144,14 +144,19 @@ jobs:
         shell: bash
         run: |
           echo "timestamp=$(git show --format='%ct' --no-patch)" >> "${GITHUB_OUTPUT}"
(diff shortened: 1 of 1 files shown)


danielocfb:
Going ahead and merging this change, to be able to monitor during working hours. I had it running in KPD for a while and everything seemed fine, but testing there only gets us so far.

PR #176, opened 2022-11-16
Topic/incr build improvs pr

danielocfb:
Assortment of improvements to incremental build logic. Please see individual commits for descriptions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -181,34 +181,39 @@ jobs:
             false
           }
 
-          # $1 - build time stamp
+          # $1 - value of KBUILD_OUTPUT
+          # $2 - current time in ISO 8601 format
           restore_source_code_times() {
-            local build_timestamp="${1}"
-            git ls-files | xargs --max-args=10000 touch -m --no-create --date="${build_timestamp}"
+            local kbuild_output="${1}"
+            local current_time="${2}"
+            local src_time="$(date --iso-8601=ns --date="${current_time} - 2 minutes")"
+            local obj_time="$(date --iso-8601=ns --date="${current_time} - 1 minute")"
+
+            git ls-files | xargs --max-args=10000 touch -m --no-create --date="${src_time}"
+            find "${kbuild_output}" -type f | xargs --max-args=10000 touch -m --no-create --date="${obj_time}"
             git checkout --quiet -
-            echo "Restored source code time stamp to time of last build"
+            echo "Adjusted src and obj time stamps relative to system time"
           }
 
           mkdir --parents "${KBUILD_OUTPUT}"
+          current_time="$(date --iso-8601=ns)"
           build_base_sha="${{ github.sha }}"
           echo "Using build base SHA-1 ${build_base_sha}"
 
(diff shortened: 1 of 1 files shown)


mykola-lysenko:
nit: "but with ISO 8601 Git would simply encode the
committer's time zone can directly be encoded" - this one reads strange, in the first commit message

danielocfb:
> nit: "but with ISO 8601 Git would simply encode the committer's time zone can directly be encoded" - this one reads strange, in the first commit message

Thanks, I fixed it up.

mykola-lysenko (approved):

mykola-lysenko:
Let's wait for tests to pass before merging

.github/workflows/test.yml
```diff
@@ -181,34 +181,39 @@ jobs:
...
 
-          # $1 - build time stamp
+          # $1 - value of KBUILD_OUTPUT
+          # $2 - current time in ISO 8601 format
           restore_source_code_times() {
-            local build_timestamp="${1}"
-            git ls-files | xargs --max-args=10000 touch -m --no-create --date="${build_timestamp}"
+            local kbuild_output="${1}"
+            local current_time="${2}"
+            local src_time="$(date --iso-8601=ns --date="${current_time} - 2 minutes")"
```
chantra, line 194: :exploding_head:

chantra (approved):

.github/workflows/test.yml
```diff
@@ -143,17 +143,22 @@ jobs:
         id: get-commit-metadata
         shell: bash
         run: |
-          echo "timestamp=$(git show --format='%ct' --no-patch)" >> "${GITHUB_OUTPUT}"
           if [ ${{ github.event_name }} = 'push' ]; then
             echo "branch=${{ github.ref_name }}" >> "${GITHUB_OUTPUT}"
+            # Ignore top most commit which just adds CI files.
+            commit="$(git rev-parse HEAD^)"
```
chantra, line 149: I wonder if we could get kpd to add a tag to the commits it generates, this way we could strip the topmost commits with the tag present.
Reasoning is that the day we add a new commit via KPD, this will break.

chantra (approved):

.github/workflows/test.yml
danielocfb: The problem I see is that then we rely on the tag being present itself and the breakage of it not being there would be more severe (`HEAD^` is always present, but it may reference a non-upstream commit or one earlier than expected; neither conditions really break anything here, as I tried to explain in the summary). We can guard against the tag not being there, but then we are back at square one on the fallback path.

Ideally we'd need neither and have a way to identify something as upstream. We could do tricks such as take the first commit with a `Link:` tag in its summary, but I felt that's not more stable either. My preference would be to use something like `git-merge-base` against a known branch, but I haven't been able to make that work -- the clone/checkout we work on didn't have a whole lot of stuff to work with. I can play around some more and see if I can improve it.

danielocfb: Going to land the change first, though, as I still think it's an improvement.

PR #179, opened 2022-12-02
selftests/bpf: Change the name kprobe_multi_test/bench_attach to kprobe_multi_bench_attach in DENYLIST.

ThinkerYzu1:
Since the test case was renamed, the DENYLIST should also be updated.

Signed-off-by: Kui-Feng Lee <kuifeng@fb.com>


danielocfb:
Thanks. Next time, please mention the upstream change that performed the rename so that people can follow along.

PR #185, opened 2022-12-29
ci/aarch64: temporarily disable entire xdp_bonding suite on aarch64

Decave:
We've been seeing frequent random failures of various testcases in the xdp_bonding suite on aarch64. A couple of other commits disabled specific testcases, but others are flakily failing as well. Let's just disable the whole suite until we investigate, as the tests aren't providing any value at the moment anyways if we're just disabling them when they fail.

Signed-off-by: David Vernet <void@manifault.com>


danielocfb (approved):
Looks good, thanks again! Will merge once CI is done.

PR #189, opened 2023-01-19
Revert "Temporarily disable testing on s390x"

danielocfb:
It appears as if the LLVM snapshot breakage for s390x has been resolved upstream and we can build and test on this architecture again. This change reverts commit 89413fe98c40f497d7f3fc6435b702ab744a9a12.


chantra (approved):
LGTM, let see how the test come back.

PR #194, opened 2023-02-13
Cross-compilation support

bjoto:
This PR adds cross-compilation and riscv64 support.

The Linux kernel has good support for cross-compilation. Adding support for cross-compilation is somewhat straightforward:

 * Detect when the build should be cross-compiled (build arch != target arch)
 * Add/populate ARCH and CROSS_COMPILE with the target arch
 * Download a cross-compilation toolchain

RISC-V is added as the first cross-compiled target, which a corresponding rootfs.

Dependencies:
 * https://github.com/libbpf/ci/pull/87
 * https://lore.kernel.org/bpf/20230210084326.1802597-1-bjorn@kernel.org/
 * TBC riscv64 config in tools/testing/selftests/bpf (patch in series)


chantra:
@bjoto FYI, we are now x-compiling s390x. I think it should work for `riscv64` too.

This was done through a combination of #262, #263, #264, and #265 .

Currently, patches 0001 and 0002 from https://github.com/kernel-patches/vmtest/pull/265/files are only in bpf-next.
Patch 0099 is not even sent to the mailing list and probably will never. This makes us run `bpftool` from the target hosts and relies on binfmt to make it work. Preferably bpftool would handle that natively.

Other note.... this setup currently assumes that we run the tests in the native target host.

chantra:
I will close this now that cross-compilation support has landed.

PR #203, opened 2023-03-09
s390x: disable parallel tests

chantra:
We are still ironing out quirks of parallel test and our s390x capacity would do better if we minimized its usage.


chantra:
successfully only generated parallel tests for non-s390x arch:
image
merging.

PR #209, opened 2023-03-24
ci: use llvm v17 for arm64 and s390x builds

anakryiko:
Seems like x86-64 is fine with llvm 16, but arm64 and s390x builds are failing right now ([0]). Try switching them back to 17, again.

  [0] https://github.com/kernel-patches/bpf/actions/runs/4508720202/jobs/7937691733


danielocfb:
Sorry, this PR did not pop up in my list. Is this something we still need? (I recall a bunch of back and forth and ultimately the root cause of those broken dependency specifications being fixed upstream)

anakryiko:
probably not, I don't remember well, but I think we tried something different at the same time, which also fixed the original issue. Will close.

PR #214, opened 2023-04-05
Move the embedded python script to an external script

chantra:
It is quite difficult to read/modify embedded scripts.
This change moves the content into an external script that can be linted/tested.

A sample test has been added. While coverage is limited, this can be used as a footstep for more testing.
Finally, python scripts are also passed through `black` as part of the CI to make sure we have consistent code style.


danielocfb (approved):
Looks good to me! Thanks!

chantra:
@yurinnick did some of this work in https://github.com/kernel-patches/vmtest/pull/233

This PR contained some testing too, which https://github.com/kernel-patches/vmtest/pull/233 does not, but given the time that as past, it does not make sense to carry on this PR, but rather build on top of https://github.com/kernel-patches/vmtest/pull/233

PR #215, opened 2023-04-13
[workflow] Add a 'veristat' job

eddyz87:
Adds a job to execute veristat on a selected set of object files as a part of the general test pipeline. The job is executed differently for pushes and for pull requests:

- For pushes:
  - the kernel is compiled as usual;
  - selftests are compiled as well, this is necessary to produce object files that would be used by veristat;
  - veristat is executed inside qemu, performance statistics is collected and cached as baseline data.

- For pull requests:
  - veristat is executed inside qemu and performance statistics is collected;
  - baseline data is restored from cache;
  - baseline data is compared to current statistics and step summary page is formed;
  - if there are some programs that verified successfully in the baseline but fail verification in current run, the step is marked as failed.

This also adds 'run_veristat' "entrypoint" for run_selftests.sh.

This entrypoint executes veristat on a set of binary files specified in `./tools/testing/selftests/bpf/veristat.cfg` file and saves output to the `/command_output/` directory within qemu image. This output is extracted from the image by `run-qemu` action.

This pull request depends on the following pull requests for libbpf/ci repository:
- https://github.com/libbpf/ci/pull/98
- https://github.com/libbpf/ci/pull/99

I tested these changes using my own fork of the https://github.com/kernel-patches/bpf repository.
For test purposes I had to do some adjustments compared to production:
- minimize set of tests;
- change 'runs_on' actions to execution on my machine;
- link to eddyz87/libbpf-ci instead of eddyz87/libbpf/ci where appropriate.

Here are example runs:
- CI run for push to 'bpf-next':
  https://github.com/eddyz87/bpf/actions/runs/4692697761

- CI run for feature branch that breaks verification for some of the files
  (note the 'Summary' page, same info is duplicated in the text log for the `veristat` job):
  https://github.com/eddyz87/bpf/actions/runs/4692808046

- CI run for feature that does not change verification results:
  https://github.com/eddyz87/bpf/actions/runs/4692963478

### DIFF
--- a/.github/scripts/veristat-compare.py
+++ b/.github/scripts/veristat-compare.py
@@ -1,0 +1,150 @@
+import sys
+import csv
+import os
+
+# File format:
+#
+# Columns:
+#  0. file_name
+#  1. prog_name
+#  2. verdict_base
+#  3. verdict_comp
+#  4. verdict_diff
+#  5. total_states_base
+#  6. total_states_comp
+#  7. total_states_diff
+#
+# Records sample:
+#  file-a,a,success,failure,MISMATCH,12,12,+0 (+0.00%)
+#  file-b,b,success,success,MATCH,67,67,+0 (+0.00%)
+
+TRESHOLD_PCT = 0
+
+HEADERS = ['file_name', 'prog_name', 'verdict_base', 'verdict_comp',
+           'verdict_diff', 'total_states_base', 'total_states_comp',
+           'total_states_diff']
+
+FILE        = 0
+PROG        = 1
+VERDICT_OLD = 2
+VERDICT_NEW = 3
+STATES_OLD  = 5
+STATES_NEW  = 6
+
+if len(sys.argv) == 2:
+    CSV_PATH = sys.argv[1]
+else:
+    print(f'Usage: {sys.argv[0]} <compare.csv>>')
+    sys.exit(1)
+
+if not os.path.exists(CSV_PATH):
+    print(f'# {CSV_PATH} does not exist, failing step')
+    sys.exit(1)
+
+def read_table(path):
+    with open(path, newline='') as file:
+        reader = csv.reader(file)
+        headers = next(reader)
+        if headers != HEADERS:
+            raise Exception(f'Unexpected table header for {path}: {headers}')
+        return list(reader)
+
+table = read_table(CSV_PATH)
+new_failures = False
+changes = False
+headers = ['File', 'Program', 'Verdict', 'States Diff (%)']
+html_table = [headers]
+text_table = [headers]
+
+def compute_diff(v):
+    old = int(v[STATES_OLD]) if v[STATES_OLD] != 'N/A' else 0
+    new = int(v[STATES_NEW]) if v[STATES_NEW] != 'N/A' else 0
+    if old == 0:
+        return 1
+    return (new - old) / old
+
+for v in table:
+    add = False
+    html_verdict = v[VERDICT_NEW]
+    text_verdict = v[VERDICT_NEW]
+    diff = compute_diff(v)
+
+    if v[VERDICT_OLD] != v[VERDICT_NEW]:
+        changes = True
+        add = True
+        html_verdict = f'{v[VERDICT_OLD]} &rarr; {v[VERDICT_NEW]}'
+        text_verdict = f'{v[VERDICT_OLD]} -> {v[VERDICT_NEW]}'
+        if v[VERDICT_NEW] == 'failure':
+            new_failures = True
+            html_verdict += ' :bangbang:'
+            text_verdict += ' (!!)'
+
+    if abs(diff * 100) > TRESHOLD_PCT:
+        changes = True
+        add = True
+
+    if not add:
+        continue
+
+    diff_txt = '{:+.1f} %'.format(diff * 100)
+    html_table.append([v[FILE], v[PROG], html_verdict, diff_txt])
+    text_table.append([v[FILE], v[PROG], text_verdict, diff_txt])
+
+def print_table(rows, fn):
+    column_widths = [0] * len(rows[0])
+    for row in rows:
+        for i, col in enumerate(row):
+            column_widths[i] = max(column_widths[i], len(col))
+
+    def print_row(row):
+        line = '|'
+        for width, col in zip(column_widths, row):
+            line += ' ' + col.ljust(width) + ' |'
+        fn(line)
+
+    rows_iter = iter(rows)
+    print_row(next(rows_iter))
+
+    underscores = '|'
+    for width in column_widths:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -353,3 +355,84 @@ jobs:
...
+          set -e
+          cd ${{ github.workspace }}
+          selftests/bpf/veristat -o csv -e file,prog,verdict,states -C \
+            veristat-baseline.csv veristat.csv > compare.csv
+          python3 ./.github/scripts/veristat-compare.py compare.csv
+
+      # For push: just put baseline log to cache
+      - if: ${{ github.event_name == 'push' }}
+        run: |
+          cp '${{ github.workspace }}/veristat.csv' \
```
danielocfb, line 431: Nit: Can this be a `mv`, actually?

ci/vmtest/run_selftests.sh
```diff
@@ -98,6 +99,17 @@ test_verifier() {
   foldable end test_verifier
 }
 
+run_veristat() {
+  foldable start run_veristat "Running veristat"
+
+  globs=$(awk '/^#/ { next; } { print $0 ".bpf.o"; }' ./veristat.cfg)
+  mkdir -p ${OUTPUT_DIR}
+  ./veristat -o csv -q -e file,prog,verdict,states ${globs} > ${OUTPUT_DIR}/veristat.csv
```
danielocfb, line 107: Can we please use long options, if available?

.github/workflows/test.yml
```diff
@@ -353,3 +355,84 @@ jobs:
...
+        with:
+          key: x86_64-gcc-veristat-baseline
+          restore-keys: |
+            x86_64-gcc-veristat-baseline-
+          path: '${{ github.workspace }}/veristat-baseline.csv'
+
+      - if: ${{ github.event_name == 'pull_request' }}
+        name: Show veristat comparison
+        run: |
+          set -e
```
danielocfb, line 422: Unnecessary? https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

.github/workflows/test.yml
```diff
@@ -295,16 +299,14 @@ jobs:
             --exclude '*.h' \
             --exclude '*.output' \
             selftests/bpf/ | zstd -T0 -19 -o vmlinux-${{ matrix.arch }}-${{ matrix.toolchain_full }}.tar.zst
-      - if: ${{ github.event_name != 'push' }}
-        name: Remove KBUILD_OUTPUT contents
+      - name: Remove KBUILD_OUTPUT contents
```
danielocfb, line 302: Can you please clarify why this change is necessary? It strikes me as wrong. Now we are no longer creating caches at all, right? I.e., it breaks incremental builds.
Also, comment in line 305 is outdated now.

eddyz87: This might be unnecessary indeed. I think I read it as '==' for some reason.
Thanks for all the comments, I'll reply in each after making changes and re-testing.

.github/scripts/veristat-compare.py
```diff
@@ -0,0 +1,150 @@
...
+    print_summary()
+    print_summary('<details>')
+    print_summary('<summary>Click to expand</summary>')
+    print_summary()
+    print_table(html_table, print_summary)
+    print_summary()
+    print_summary('</details>')
+
+if new_failures:
+    print(f'Failing step because of the new failures')
```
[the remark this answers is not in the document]
eddyz87: Removed

.github/workflows/test.yml
```diff
@@ -353,3 +355,84 @@ jobs:
...
+    env:
+      KERNEL: LATEST
+      REPO_ROOT: ${{ github.workspace }}
+      REPO_PATH: ""
+      KBUILD_OUTPUT: kbuild-output/
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/download-artifact@v3
+        with:
+          name: vmlinux-x86_64-gcc
```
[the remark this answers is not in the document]
eddyz87: Added to env variables at the top level:

```yaml
env:
  veristat_arch: x86_64
  veristat_toolchain: gcc
```

And a step to derive the prefix at the start of the veristat job:

```yaml
      - name: Setup environment variables
        run: |
            echo arch_and_tool=${{ env.veristat_arch }}-${{ env.veristat_toolchain }} > \
              ${GITHUB_ENV}
```

Please let me know if there is a more elegant way to do it, extending `env` section at the job level to introduce the var using top-level `env.veristat_*` values didn't work for some reason.

Also changed `set-matrix` code that computes `veristat_runs_on` to use the above variables.

.github/workflows/test.yml
eddyz87: Changed to `mv`.

ci/vmtest/run_selftests.sh
eddyz87: Done

.github/workflows/test.yml
eddyz87: Unnecessary indeed, removed.

.github/workflows/test.yml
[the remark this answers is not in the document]
danielocfb: What you do seems fine, thanks!

danielocfb (approved):
Looks good to me now, thanks! Will merge tomorrow, unless Manu wants to take a look as well.

danielocfb:
We are seeing failures in CI, e.g., https://github.com/kernel-patches/bpf/actions/runs/4756853058/jobs/8453383397

Will revert the change for the time being.

eddyz87:
It complains that veristat baseline is not present. The base line is generated after CI execution for a push to bpf-next. So, it won't be available for some time. Is this pipeline for pull request of a push? I can relax the condition, allowing this step to not fail if there is no baseline.

danielocfb:
Correct, yes, we should not error out if it's not present, I would say. In general we expect to have a baseline available (just by virtue of how frequently we see pushes), but we cannot guarantee that. E.g., we've had long stretches in the past where no changes got pushed and it's conceivable that GitHub Actions evicts cached artifacts (we have limited control over that).

danielocfb:
> Is this pipeline for pull request of a push?

This is the job for the push: https://github.com/kernel-patches/bpf/actions/runs/4756852477/jobs/8453428926
It did succeed. But let's fix this up.

eddyz87:
Ok, will do that. I need to create new merge request for that, right?
I'll post an update in ~2hr, need to wrap up another task + local testing takes some time.

danielocfb:
> Ok, will do that. I need to create new merge request for that, right?

Yeah, please do that. Will probably get merged only next week then, because I'll be out tomorrow.

PR #218, opened 2023-04-20
[workflow] Add a 'veristat' job

eddyz87:
Adds a job to execute veristat on a selected set of object files as a part of the general test pipeline. The job is executed differently for pushes and for pull requests:

- For pushes:
  - the kernel is compiled as usual;
  - selftests are compiled as well, this is necessary to produce object files that would be used by veristat;
  - veristat is executed inside qemu, performance statistics is collected and cached as baseline data.

- For pull requests:
  - veristat is executed inside qemu and performance statistics is collected;
  - baseline data is restored from cache;
  - baseline data is compared to current statistics and step summary page is formed;
  - if there are some programs that verified successfully in the baseline but fail verification in current run, the step is marked as failed.

This also adds 'run_veristat' "entrypoint" for run_selftests.sh.

This entrypoint executes veristat on a set of binary files specified in ./tools/testing/selftests/bpf/veristat.cfg file and saves ouput to the /command_output/ directory within qemu image. This output is extracted from the image by run-qemu action.

This is a followup for [this](https://github.com/kernel-patches/vmtest/pull/215) pull request, which was reverted. This one is changed to not fail `veristat` job if baseline is not available. Here is a link to a test execution under such conditions:

https://github.com/eddyz87/bpf/actions/runs/4758707630/jobs/8457542575

### DIFF
--- a/.github/scripts/veristat-compare.py
+++ b/.github/scripts/veristat-compare.py
@@ -1,0 +1,202 @@
+#!/usr/bin/env python3
+
+# This script reads a CSV file produced by the following invocation:
+#
+#   veristat --emit file,prog,verdict,states \
+#            --output-format csv \
+#            --compare ...
+#
+# And produces a markdown summary for the file.
+# The summary is printed to standard output and appended to a file
+# pointed to by GITHUB_STEP_SUMMARY variable.
+#
+# Script exits with return code 1 if there are new failures in the
+# veristat results.
+#
+# For testing purposes invoke as follows:
+#
+#  GITHUB_STEP_SUMMARY=/dev/null python3 veristat-compare.py test.csv
+#
+# File format (columns):
+#  0. file_name
+#  1. prog_name
+#  2. verdict_base
+#  3. verdict_comp
+#  4. verdict_diff
+#  5. total_states_base
+#  6. total_states_comp
+#  7. total_states_diff
+#
+# Records sample:
+#  file-a,a,success,failure,MISMATCH,12,12,+0 (+0.00%)
+#  file-b,b,success,success,MATCH,67,67,+0 (+0.00%)
+#
+# For better readability suffixes '_OLD' and '_NEW'
+# are used instead of '_base' and '_comp' for variable
+# names etc.
+
+import io
+import os
+import sys
+import csv
+import logging
+import argparse
+from functools import reduce
+from dataclasses import dataclass
+
+TRESHOLD_PCT = 0
+
+HEADERS = ['file_name', 'prog_name', 'verdict_base', 'verdict_comp',
+           'verdict_diff', 'total_states_base', 'total_states_comp',
+           'total_states_diff']
+
+FILE        = 0
+PROG        = 1
+VERDICT_OLD = 2
+VERDICT_NEW = 3
+STATES_OLD  = 5
+STATES_NEW  = 6
+
+# Given a table row, compute relative increase in the number of
+# processed states.
+def compute_diff(v):
+    old = int(v[STATES_OLD]) if v[STATES_OLD] != 'N/A' else 0
+    new = int(v[STATES_NEW]) if v[STATES_NEW] != 'N/A' else 0
+    if old == 0:
+        return 1
+    return (new - old) / old
+
+@dataclass
+class VeristatInfo:
+    table: list
+    changes: bool
+    new_failures: bool
+
+# Read CSV table expecting the above described format.
+# Return VeristatInfo instance.
+def parse_table(csv_filename):
+    new_failures = False
+    changes = False
+    table = []
+
+    with open(csv_filename, newline='') as file:
+        reader = csv.reader(file)
+        headers = next(reader)
+        if headers != HEADERS:
+            raise Exception(f'Unexpected table header for {filename}: {headers}')
+
+        for v in reader:
+            add = False
+            verdict = v[VERDICT_NEW]
+            diff = compute_diff(v)
+
+            if v[VERDICT_OLD] != v[VERDICT_NEW]:
+                changes = True
+                add = True
+                verdict = f'{v[VERDICT_OLD]} -> {v[VERDICT_NEW]}'
+                if v[VERDICT_NEW] == 'failure':
+                    new_failures = True
+                    verdict += ' (!!)'
+
+            if abs(diff * 100) > TRESHOLD_PCT:
+                changes = True
+                add = True
+
+            if not add:
+                continue
+
+            diff_txt = '{:+.1f} %'.format(diff * 100)
+            table.append([v[FILE], v[PROG], verdict, diff_txt])
+
+    return VeristatInfo(table=table,
+                        changes=changes,
+                        new_failures=new_failures)
+
+def format_table(headers, rows, html_mode):
+    def decorate(val, width):
+        s = str(val)
+        if html_mode:
+            s = s.replace(' -> ', ' &rarr; ');
+            s = s.replace(' (!!)', ' :bangbang: ');
+        return s.ljust(width)
+
+    column_widths = list(reduce(lambda acc, row: map(max, map(len, row), acc),
+                                rows,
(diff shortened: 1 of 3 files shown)


danielocfb (approved):
Looks good to me, thanks Eduard!

.github/scripts/veristat-compare.py
```diff
@@ -0,0 +1,150 @@
...
+STATES_NEW  = 6
+
+if len(sys.argv) == 2:
+    CSV_PATH = sys.argv[1]
+else:
+    print(f'Usage: {sys.argv[0]} <compare.csv>>')
+    sys.exit(1)
+
+if not os.path.exists(CSV_PATH):
+    print(f'# {CSV_PATH} does not exist, failing step')
```
chantra, line 41: any reasons to print to stdout vs stderr? or not using the logging module?

.github/scripts/veristat-compare.py
```diff
@@ -0,0 +1,150 @@
...
+        if headers != HEADERS:
+            raise Exception(f'Unexpected table header for {path}: {headers}')
+        return list(reader)
+
+table = read_table(CSV_PATH)
+new_failures = False
+changes = False
+headers = ['File', 'Program', 'Verdict', 'States Diff (%)']
+html_table = [headers]
+text_table = [headers]
```
chantra, line 57: looks like those are global variables. Could we limit the use of globals and favor passing arguments or objects to abstract it?

.github/scripts/veristat-compare.py
```diff
@@ -0,0 +1,150 @@
...
+HEADERS = ['file_name', 'prog_name', 'verdict_base', 'verdict_comp',
+           'verdict_diff', 'total_states_base', 'total_states_comp',
+           'total_states_diff']
+
+FILE        = 0
+PROG        = 1
+VERDICT_OLD = 2
+VERDICT_NEW = 3
+STATES_OLD  = 5
+STATES_NEW  = 6
```
chantra, line 32: This could be wrapped in an [enum](https://docs.python.org/3/library/enum.html)

.github/scripts/veristat-compare.py
```diff
@@ -0,0 +1,150 @@
...
+
+HEADERS = ['file_name', 'prog_name', 'verdict_base', 'verdict_comp',
+           'verdict_diff', 'total_states_base', 'total_states_comp',
+           'total_states_diff']
+
+FILE        = 0
+PROG        = 1
+VERDICT_OLD = 2
+VERDICT_NEW = 3
+STATES_OLD  = 5
```
chantra, line 31: is 4 missing on purpose?

danielocfb:
Manu mentioned that he also wants to take a look, so will wait for him.

.github/scripts/veristat-compare.py
```diff
@@ -0,0 +1,150 @@
...
+if new_failures:
+    section_name = '# There are new veristat failures'
+elif changes:
+    section_name = '# There are changes in verification performance'
+else:
+    section_name = '# No changes in verification performance'
+
+# Print the step summary
+
+if summary_path := os.getenv('GITHUB_STEP_SUMMARY', None):
```
chantra, line 125: TIL `:=` :D
This was introduced in python3.8 apparently. The containers have python3.8.10 so we should be fine with this syntax.

I am not sure we should silently ignore this. This script's main purpose is to provide this summary, right?

When running outside of GH actions, this can be worked around by defining the env var on the command line.

For #214 I basically attempt to document this at the top of the file with an example run one-liner: https://github.com/kernel-patches/vmtest/pull/214/files#diff-cf4f5f0c0e0169cd40d613cf183cf3d4a8b8f058e2560b3853ef6aa3717dcdfaR3

This file could probably benefit from this too?

chantra (commented):
Thanks @eddyz87 !

One comment I have regarding the python script is that currently it is using `globals`, which I would rather avoid and explicitely pass parameters around.

Another thing is that it seems to mix program control flow with definitions of functions.
Also, I would rather have the main control flow gated under `if __name__ == '__main__'`.

I would favor use of argparse to manually dealing with arguments. While probably not a big deal now, it will make future changes easier, and we don't have to deal with arg parsing when it is one import away in core python.
Same for logging, let's use the logging module.
Those 2 are the canonical way to handle args, log in python, let's re-use them unless there is a good reason not to.

I think this change could benefit from shipping fixture for cases where we expect the script to report failure and when we expect it to be fine. There again, in the vain of #214, we would be able to validate:
1) we exist with the return code we expect (regression detected or not)
2) We have a simple way to confirm that the script does not explode at run time.

While not necessarily asking to hook the testing in this PR, having those fixtures, they could be used as part of the test plan.

chantra:
TL;DR-ing my comments. This is non-blocking, but I think addressing them will make the project more sustainable. We have already accumulated a bunch of tech-debts in the past and this is making the project hard to maintain, so I would like to avoid adding more to it.

Are you onboard for addressing those comments in future diffs?

.github/scripts/veristat-compare.py
```diff
@@ -0,0 +1,150 @@
+import sys
```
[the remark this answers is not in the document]
eddyz87: Will do

.github/scripts/veristat-compare.py
```diff
@@ -0,0 +1,150 @@
...
+VERDICT_OLD = 2
+VERDICT_NEW = 3
+STATES_OLD  = 5
+STATES_NEW  = 6
+
+if len(sys.argv) == 2:
+    CSV_PATH = sys.argv[1]
+else:
+    print(f'Usage: {sys.argv[0]} <compare.csv>>')
+    sys.exit(1)
```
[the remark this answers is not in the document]
eddyz87: Will do

.github/scripts/veristat-compare.py
eddyz87: No reasons, just wanted to keep things simple, will change to `logging`

.github/scripts/veristat-compare.py
eddyz87: will wrap as a function

.github/scripts/veristat-compare.py
eddyz87: These are used as indices in a vector, using `Enum` would make the notation less readable: `row[FILE]` &rarr; `row[FILE.Value]`, so I'd rather not use enum in this case.

.github/scripts/veristat-compare.py
eddyz87: yes, sample of the record in the comment at the [start](https://github.com/kernel-patches/vmtest/pull/218/commits/88630c9ea660724f5017584bb260ea63645a7766#diff-d9bba8d01ee121eade3ef637cc83865fac60cb310a246ae82de3fd98c6dfd33eR5) of the file, here are the columns:

```
# Columns:
#  0. file_name
#  1. prog_name
#  2. verdict_base
#  3. verdict_comp
#  4. verdict_diff
#  5. total_states_base
#  6. total_states_comp
#  7. total_states_diff
```

The `verdict_diff` is ignored.

.github/scripts/veristat-compare.py
eddyz87: I can skip the conditional, will rely on python exiting with error if file can't be opened.

eddyz87:
> TL;DR-ing my comments. This is non-blocking, but I think addressing them will make the project more sustainable. We have already accumulated a bunch of tech-debts in the past and this is making the project hard to maintain, so I would like to avoid adding more to it.
> 
> Are you onboard for addressing those comments in future diffs?

I can address these today as a part of this pull request.

chantra:
Thanks @eddyz87 !

eddyz87:
> Thanks @eddyz87 !

Just pushed the updated script, I think I covered all the above points.

chantra:
Thanks @eddyz87 . Sorry for the late review. Let's merge this tomorrow.

PR #219, opened 2023-04-22
run set-matrix on self-hosted runners for kernel-patches org

chantra:
With bpf-rc repo using GH hosted runners, we have seen cases when bpf actions are waiting on a GH runner. An org is limited to 20 GH runners, when multiple series are being processed, we max them out and bpf repo, which predominantly runs on self-hosted runners, is being idle waiting for a GH runner to free up, perform the action in a couple of seconds and move on.

This change will run set-matrix on any self-hosted machine free and ready to take an action. It does not need any specific arch, so let's pick whatever.
This mean that we will possibly run set-matrix for repo that are not running off self-hosted, on self-hosted, but that should not be a problem. in term of resources.

Here the trick is to evaluate the runner to use as part of a github action expresion so it is executed by github, and not the runner.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,8 +12,14 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+
   set-matrix:
-    runs-on: ubuntu-latest
+    # FIXME: set-matrix is lightweight, run it on any self-hosted machines for kernel-patches org
+    # so we do not wait for GH hosted runners when there potentially all are busy because of bpf-rc
+    # repo for instance.
+    # This could be somehow fixed long term by making this action/workflow re-usable and letting the called
+    # specify what to run on.
+    runs-on: ${{ github.repository_owner == 'kernel-patches' && 'self-hosted' || 'ubuntu-latest' }}
     outputs:
       build-matrix: ${{ steps.set-matrix-impl.outputs.build_matrix }}
       test-matrix: ${{ steps.set-matrix-impl.outputs.test_matrix }}


.github/workflows/test.yml
```diff
@@ -12,8 +12,11 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+
   set-matrix:
-    runs-on: ubuntu-latest
+    # FIXME: set-matrix is lightweight, run it on any self-hosted machines for kernel-patches so we do
```
danielocfb, line 17: It's not really clear to me what is there to fix here, can you make that a bit more clear?

danielocfb (commented):
So if I am understanding correctly, this will make this job *only* eligible to run on `self-hosted`, correct (for `kernel-patches` org)? But why restrict usage of GitHub hosted when their may be free ones? It feels like we are mostly shifting the issue from one pool to another. Shouldn't GitHub Actions be able to use any available runner in either pool? I'd think that can be accomplished with syntax such as this one: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-labels-for-runner-selection

Have you considered that or am I misunderstanding?

chantra:
> So if I am understanding correctly, this will make this job _only_ eligible to run on `self-hosted`, correct (for `kernel-patches` org)?

yes

> But why restrict usage of GitHub hosted when their may be free ones? It feels like we are mostly shifting the issue from one pool to another. Shouldn't GitHub Actions be able to use any available runner in either pool? I'd think that can be accomplished with syntax such as this one: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-labels-for-runner-selection
> 
> Have you considered that or am I misunderstanding?

Yes, my understanding is that all labels are AND-ed together so I can't really specify self-hosted + GH-hosted.  The gh-hosted only have the label ubuntu-latest, so I can't really find a tag that intersect. In any cases, this job is very lightweight and very unlikely to be our bottleneck for now. So I think it is a fine trade-off.

>  When you specify an array of labels, jobs will be queued on runners that have all the labels that you specify.

.github/workflows/test.yml
chantra: oh boy, I must have written this comment without thinking it through.

I think I marked this as a FIXME because once we refactor this workflow/job to be re-usable, we can differ setting on what runner to run on to the caller (and default to ubuntu-latest). Will make it more explicit here.

danielocfb (approved):
Thanks for clarifying! Seems fine to me.

PR #236, opened 2023-09-13
Fix veristat baseline cache save/restore

yurinnick:
Github cache has S3-like structure, where key is a bucket and files specified in the path argument makes its content.
Due to that fact save and restore actions MUST have the same path argument.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -243,11 +243,11 @@ jobs:
       - name: Compare and save veristat.kernel.csv
         uses: ./.github/actions/veristat_baseline_compare
         with:
-          arch_and_tool: ${{ env.ARCH_AND_TOOL}}
           veristat_output: veristat-kernel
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -243,11 +243,13 @@ jobs:
...
+          baseline_name: ${{ env.ARCH_AND_TOOL}}-baseline-veristat-kernel
 
       - name: Compare and save veristat.meta.csv
         uses: ./.github/actions/veristat_baseline_compare
         with:
-          arch_and_tool: ${{ env.ARCH_AND_TOOL}}
           veristat_output: veristat-meta
+          baseline_name: ${{ env.ARCH_AND_TOOL}}-baseline-veristat-meta
+
+
```
chantra, line 255: You could remove those before the next auto-formatting editor does it.

chantra (approved):
let's wait for the tests to confirm this is working as expected. Otherwise, LGTM

PR #237, opened 2023-10-04
Refactor veristat-compare.py script

yurinnick:
- Use csv.DictReader for veristat results parsing
- Better column name handling with Enum class
- Fix the issue when comparison reports 100% change when both old and new values are zeros

### DIFF
--- a/.github/scripts/veristat-compare.py
+++ b/.github/scripts/veristat-compare.py
@@ -38,33 +38,71 @@
 import io
 import os
 import sys
+import re
 import csv
 import logging
 import argparse
-from functools import reduce
+import enum
 from dataclasses import dataclass
+from typing import Dict, List, Final
 
-TRESHOLD_PCT = 0
 
-HEADERS = ['file_name', 'prog_name', 'verdict_base', 'verdict_comp',
-           'verdict_diff', 'total_states_base', 'total_states_comp',
-           'total_states_diff']
+TRESHOLD_PCT: Final[int] = 0
 
-FILE        = 0
-PROG        = 1
-VERDICT_OLD = 2
-VERDICT_NEW = 3
-STATES_OLD  = 5
-STATES_NEW  = 6
+SUMMARY_HEADERS = ["File", "Program", "Verdict", "States Diff (%)"]
 
-# Given a table row, compute relative increase in the number of
-# processed states.
-def compute_diff(v):
-    old = int(v[STATES_OLD]) if v[STATES_OLD] != 'N/A' else 0
-    new = int(v[STATES_NEW]) if v[STATES_NEW] != 'N/A' else 0
-    if old == 0:
-        return 1
-    return (new - old) / old
+# expected format: +0 (+0.00%) / -0 (-0.00%)
+TOTAL_STATES_DIFF_REGEX = \
+    r"(?P<absolute_diff>[+-]\d+) \((?P<percentage_diff>[+-]\d+\.\d+)\%\)"
+
+
+TEXT_SUMMARY_TEMPLATE: Final[str] = """
+# {title}
+
+{table}
+""".strip()
+
+HTML_SUMMARY_TEMPLATE: Final[str] = """
+# {title}
+
+<details>
+<summary>Click to expand</summary>
+
+{table}
+</details>
+""".strip()
+
+GITHUB_MARKUP_REPLACEMENTS: Final[Dict[str, str]] = {
+    "->": "&rarr;",
+    "(!!)": ":bangbang:",
+}
+
+NEW_FAILURE_SUFFIX: Final[str] = "(!!)"
+
+
+class VeristatFields(str, enum.Enum):
+    FILE_NAME = "file_name"
+    PROG_NAME = "prog_name"
+    VERDICT_OLD = "verdict_base"
+    VERDICT_NEW = "verdict_comp"
+    VERDICT_DIFF = "verdict_diff"
+    TOTAL_STATES_OLD = "total_states_base"
+    TOTAL_STATES_NEW = "total_states_comp"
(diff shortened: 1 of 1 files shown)


.github/scripts/veristat-compare.py
```diff
@@ -38,165 +38,212 @@
...
+        return [
+            cls.FILE_NAME,
+            cls.PROG_NAME,
+            cls.VERDICT_OLD,
+            cls.VERDICT_NEW,
+            cls.VERDICT_DIFF,
+            cls.TOTAL_STATES_OLD,
+            cls.TOTAL_STATES_NEW,
+            cls.TOTAL_STATES_DIFF,
+        ]
```
chantra, line 97: if those are all the members:
```
return [x for x in cls]

File: .github/scripts/veristat-compare.py [hunk 2]
```diff
@@ -38,165 +38,212 @@
...
+
+        if self.changes:
+            return "There are changes in verification performance"
+
+        return "No changes in verification performance"
+
+    def get_results_summary(self, html: bool = False) -> str:
+        title = self.get_results_title()
+        if not self.table:
+            return f"# {title}\n"
```
chantra, line 118: as a follow-up, it would be good to report failure/warning with color codes so it is made more obvious in GH.

.github/scripts/veristat-compare.py
```diff
@@ -38,165 +38,212 @@
...
 
-        out.write('|')
-        out.write('|'.join(map(lambda w: '-' * (w + 2), column_widths)))
-        out.write('|\n')
 
-        for row in rows:
-            print_row(row)
+def html_decorate(input_str: str) -> str:
+    return input_str\
+        .replace(" -> ", " &rarr; ")\
```
chantra, line 190: https://docs.python.org/3/library/html.html#html.escape ?

.github/scripts/veristat-compare.py
```diff
@@ -38,165 +38,212 @@
...
-            out.write('| ')
-            out.write(' | '.join(map(decorate, row, column_widths)))
-            out.write(' |\n')
-
-        print_row(headers)
+    for record in reader:
+        add = False
+
+        verdict_old, verdict_new = (
+            record[VeristatFields.VERDICT_OLD],
```
chantra, line 152: should you use `.value`?

```
>>> VeristatFields.FILE_NAME.value
'file_name'
```

.github/scripts/veristat-compare.py
```diff
@@ -38,165 +38,212 @@
...
+            record[VeristatFields.VERDICT_NEW],
+        )
+
+        if record[VeristatFields.VERDICT_DIFF] == "MISMATCH":
+            changes = True
+            add = True
+            verdict = f"{verdict_old} -> {verdict_new}"
+            if verdict_new == "failure":
+                new_failures = True
+                verdict += " (!!)"
```
chantra, line 162: I feel this `(!!)` should be a constant.

chantra (approved):
Skimmed over and looks fine.

It would be useful to capture some fixtures and run tests against those to make sure we don't break the code in the future as it is hard to really validate without going through a full cycle.

The veristat summary of this diff is:
```
# No /tmp/work/vmtest/vmtest/x86_64-gcc-baseline-veristat-kernel available

# No /tmp/work/vmtest/vmtest/x86_64-gcc-baseline-veristat-meta available
```
is this expected?

.github/scripts/veristat-compare.py
yurinnick: I am concerned about the order on the elements, because it's important

.github/scripts/veristat-compare.py
yurinnick: Github markup doesn't support colors: https://github.com/github/markup/issues/1440

.github/scripts/veristat-compare.py
yurinnick: It escapes html tags, but doesn't replace text with a nicer symbols

.github/scripts/veristat-compare.py
yurinnick: Yes, if you want to print it. But we add these into f-strings which calls `__str__()`. `VeristatFields(str, Enum)`, so `__str__()` for it will return `str` value of the enum by default

yurinnick:
> is this expected?

Yes, since we don't store any baseline metrics for in this repository`s cache, we don't have anything to compare to

.github/scripts/veristat-compare.py
chantra: > It escapes html tags, but doesn't replace text with a nicer symbols

ha yeah.... `->` would have been translated to `-&gt;` instead.

.github/scripts/veristat-compare.py
chantra: The text content can be "colorized" by mean of special tags, see https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message

The markup one can use simple to signal success/failure: https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/

chantra:
> > is this expected?
> 
> Yes, since we don't store any baseline metrics for in this repository`s cache, we don't have anything to compare to

ok, I guess there is still value to run it in vmtest even without the baseline. So, let's leave it like this.


Did you test this in kernel-patches/bpf? Could you make a PR where vmtest is modified to reflect the changes you are making here so we can have a proper e2e test?

yurinnick:
Created test PR in `kernel-patches/bpf`: kernel-patches/bpf#5781

.github/scripts/veristat-compare.py
yurinnick: It's not exactly "colors", but I got what you mean. I will cover it in a separate PR.

PR #238, opened 2023-10-04
Skip veristat validation with Meta objects for PRs from forks

yurinnick:
Github actions doesn't allow to access both variable and secrets for workflows triggered by pull requests from forked repositories.

Disabling veristat validation with Meta objects for such PRs since workflow won't be able to download objects due to missing secrets.

Example workflow run: https://github.com/kernel-patches/bpf/actions/runs/6412314917/job/17409487960

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -209,6 +209,10 @@ jobs:
           image-output: '/tmp/root.img'
 
       - name: Configure AWS Credentials
+        # Disabling BPF objects download and veristat-meta benchmark for PRs
+        # created from fork repositories. These won't have access to required
+        # enviroment variables and secrets, and otherwise would consistently fail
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -209,13 +209,15 @@ jobs:
           image-output: '/tmp/root.img'
 
       - name: Configure AWS Credentials
+        if: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
```
chantra, line 212: Maybe leave a quick comment as to why this is gated.

chantra (approved):
LGTM, let's wait for the tests to finish and confirm it is all good.

.github/workflows/test.yml
yurinnick: Added comment

Issue #240, opened 2023-10-19
veristat-compare.py does not react well to new programs in the log


eddyz87:
Consider the following veristat input files:

`two-ok`:

```
file_name,prog_name,verdict,total_states
file-a,a,success,12
file-b,b,success,67
```

`add-success`:

```
file_name,prog_name,verdict,total_states
file-a,a,success,12
file-b,b,success,67
file-b,c,success,32
```

veristat produces the following log (`veristat -e file,prog,verdict,states -o csv -C two-ok add-success > $tmp`):

```
file_name,prog_name,verdict_base,verdict_comp,verdict_diff,total_states_base,total_states_comp,total_states_diff
file-a,a,success,success,MATCH,12,12,+0 (+0.00%)
file-b,b,success,success,MATCH,67,67,+0 (+0.00%)
file-b,c,N/A,success,N/A,N/A,32,N/A
```

And veristat-compare.py fails with the log below:

```
Traceback (most recent call last):
  File "/home/eddy/work/vmtest/.github/scripts/veristat-compare.py", line 256, in <module>
    sys.exit(main(args.filename, summary_filename))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/eddy/work/vmtest/.github/scripts/veristat-compare.py", line 233, in main
    veristat_results = parse_table(csv_file)
                       ^^^^^^^^^^^^^^^^^^^^^
  File "/home/eddy/work/vmtest/.github/scripts/veristat-compare.py", line 173, in parse_table
    diff = get_state_diff(record[VeristatFields.TOTAL_STATES_DIFF])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/eddy/work/vmtest/.github/scripts/veristat-compare.py", line 139, in get_state_diff
    if percentage_diff := matches.group("percentage_diff"):
                          ^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'group'
```

This causes CI failures, e.g. here:
https://github.com/kernel-patches/bpf/actions/runs/6567849918/job/17845229398#step:35:111

eddyz87:
[veristat-compare-testing.zip](https://github.com/kernel-patches/vmtest/files/13044442/veristat-compare-testing.zip)

I had a small "testsuite" when I initially developed the script.
It should probably be integrated as a local CI run.
[...]

PR #241, opened 2023-10-19
Fix veristat comparison for completely new and removed bpf objects

yurinnick:
New and removed objects have `N/A` value in verdicts fields, ignoring these records.

Add additional safeguard for `verdict_diff` parsing of `N/A` value.

### DIFF
--- a/.github/scripts/veristat-compare.py
+++ b/.github/scripts/veristat-compare.py
@@ -135,7 +135,15 @@ class VeristatInfo:
 
 
 def get_state_diff(value: str) -> float:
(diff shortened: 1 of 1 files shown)


chantra:
@yurinnick did you test against the example from @eddyz87 in #240 ?

How is the newly raised exception handled?

chantra:
Can you follow up after fixing the issue and capture fixtures so we can start adding e2e testing.

chantra:
veristat test is successful. Let's merge this.

PR #242, opened 2023-10-19
Add unittests for veristat_compare script

yurinnick:
- Rename veristat-compare.py into veristat_compare.py to match python naming convention
- Add unittests for parse_tables() method

### DIFF
--- a/.github/scripts/tests/test_veristat_compare.py
+++ b/.github/scripts/tests/test_veristat_compare.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python3
+
+import unittest
+from typing import Iterable, List
+
+from ..veristat_compare import parse_table, VeristatFields
+
+
+def gen_csv_table(records: Iterable[str]) -> List[str]:
+    return [
+        ",".join(VeristatFields.headers()),
+        *records,
+    ]
+
+
+class TestVeristatCompare(unittest.TestCase):
+    def test_parse_table_ignore_new_prog(self):
+        table = gen_csv_table(
+            [
+                "prog_file.bpf.o,prog_name,N/A,success,N/A,N/A,1,N/A",
+            ]
+        )
+        veristat_info = parse_table(table)
+        self.assertEqual(veristat_info.table, [])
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -16,7 +16,18 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  test-scripts:
```
chantra, line 19: this will be better in the lint.yml workflow, so we don't pollute our main workflow with skipped jobs as in https://github.com/kernel-patches/bpf/actions/runs/6582407583

.github/workflows/test.yml
```diff
@@ -16,7 +16,18 @@ concurrency:
...
+    if: ${{ github.repository == 'kernel-patches/vmtest' }}
+    name: Lint and Test python scripts
```
chantra, line 21: nit: "Test python scripts"

.github/scripts/tests/test_veristat_compare.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def gen_csv_table(records: Iterable[str]) -> List[str]:
+    return [
+        ",".join(VeristatFields.headers()),
+        *records,
+    ]
+
+
+class TestMatrix(unittest.TestCase):
```
chantra, line 16: not sure if TestMatrix is the most descriptive name for this test class. I believe this is the remnant of some copy/paste.

chantra (approved):
I have no knowledge of what the csv files look like, so can't comment. But thanks for adding the unittests.

Maybe @eddyz87 can comment?

Other than some nit: LGTM

eddyz87:
> I have no knowledge of what the csv files look like, so can't comment. But thanks for adding the unittests.
> 
> Maybe @eddyz87 can comment?

Looks good.

.github/scripts/tests/test_veristat_compare.py
yurinnick: Renamed into `TestVeristatCompare`

PR #243, opened 2023-10-21
vmtest: use danobi/vmtest instead of homegrown vm solution

chantra:
https://github.com/danobi/vmtest make it easy to run a VM with a given kernel
and using the host rootfs (RO) and a shared directory (RW).

This will help us with a couple of issue:
- we don't have to maintain a rootfs anymore
- when running the test, we do not need to perform a dance with libguestfs, which allows us to save time during test, and make it trivial to exchange files between the host and the guest.
- we won't have anymore discrepency between the libraries used in the host and guest (like explained in libbpf/ci#84 and also libbpf/ci#103)

This change migrates the tests and veristat runs from using `prepare-rootfs` + `run-qemu` actions to using `run-vmtest` instead


chantra:
Some of the testing can only be done from kernel-patches/bpf repo. This is being done in https://github.com/kernel-patches/bpf/pull/6325 / https://github.com/kernel-patches/bpf/actions/runs/7661982374

PR #244, opened 2023-10-24
workflow: move build operations in a reusable workflow

chantra:
This change essentially isolated the old jobs:
- build
- test
- veristat

in re-usable workflows: `.github/workflows/kernel-{build,test,veristat}.yml`.

There is then another reusable workflow: `.github/workflows/kernel-build-test.yml` which runs the jobs to build, test, and run veristat for a specific architecture/toolchain.

Finally, `.github/workflows/test.yml` is still the main workflow, but now run the `set-matrix` job and initiate a `kernel-build-test` workflow for each arch/toolchain pairs.
This allow us to essentially start those workflow in parallel, as a build in a specific workflow is finished, it will process with testing directly without waiting for other jobs.

Finally, matrix.py was changed to include the tests to run directly as part of the `build-matrix`. This allow us to not have to do any filtering as part of the GH actions.
In the future, the test-matrix could be removed. In the future, we may also want to embed whether or not we run veristat at the build matrix level.... which is not a build-matrix anymore, but an actual full end-to-end matrix.


chantra:
I have ran KPD against this and it seems that it will handle updating KPD just fine.

I am also opening this PR against a different branch than master so we can switch KPD to use the new vmtest code and possibly revert back to main which we know is stable.

.github/workflows/test.yml
```diff
@@ -32,229 +35,24 @@ jobs:
...
-    timeout-minutes: 100
-    env:
-      KERNEL: LATEST
-      REPO_ROOT: ${{ github.workspace }}
-      REPO_PATH: ""
-      KBUILD_OUTPUT: kbuild-output/
+
+  build-and-test:
+    uses: ./.github/workflows/kernel-build-test.yml
+    needs: [set-matrix]
```
yurinnick, line 41:
```suggestion
  build-and-test:
    name: "{{ matrix.arch }}-{{ matrix.toolchain_full }}"
    uses: ./.github/workflows/kernel-build-test.yml
    needs: [set-matrix]
```

It would render job name in a nicer way

chantra: done! thanks

.github/workflows/kernel-test.yml
```diff
@@ -0,0 +1,70 @@
...
+      continue_on_error:
+        required: true
+        type: string
+      timeout_minutes:
+        required: true
+        type: number
+
+jobs:
+  test:
+    if: ${{ github.event_name != 'push' }}
```
danielocfb, line 37: This condition doesn't really here, does it? It seems more like a property that the caller imposes, no?

.github/workflows/kernel-build.yml
```diff
@@ -0,0 +1,118 @@
...
+        REPO_ROOT: ${{ github.workspace }}
+        REPO_PATH: ""
+        KBUILD_OUTPUT: kbuild-output/
+    steps:
+      - uses: actions/checkout@v3
+        # We fetch an actual bit of history here to facilitate incremental
+        # builds (which may check out some earlier upstream change).
+        with:
+          fetch-depth: 50
+      - if: ${{ github.repository == 'kernel-patches/vmtest' }}
```
danielocfb, line 49: I'd rather we make this some kind of meaningful input argument than this reliance on global state. Do you think that's possible?

danielocfb (approved):
Overall this looks good to me, thanks! I didn't look closely at the Python logic changes.

.github/workflows/kernel-test.yml
chantra: yeah, fair point, this should be moved up one level.

chantra (commented):
Talked offline and agreed to address this separately once the bulk of the change is on master.

PR #246, opened 2023-10-27
tests: Control execution of test job from the caller, not the callee

chantra:
@danielocfb brought this up in #244.
e.g we should control whether to execute the test job not from the job itself, but from the caller.


danielocfb (approved):
Looks good to me, thanks!

PR #249, opened 2023-10-28
ci/diffs: Disable llvm feature detection for bpftool

chantra:
When selftests run and shellout to bpftool in the VM, llvm is missing and we are failing tests.


chantra:
tests are passing for x86.... landing.

PR #252, opened 2023-11-12
Add veristat.cfg for meta BPF objs

davemarchevsky:
This PR adds a `veristat_meta.cfg` to this repository, meant to be used with the tarball of meta bpf objs.

`run_selftests.sh` is modified to use veristat.cfg files via veristat's `-f` flag. See individual commit summaries for more detail.


anakryiko:
Great, thank you!

PR #254, opened 2023-12-11
ci: bump minimum clang to 17 and add 18 to x86-64

anakryiko:
Use clang-17 now that it is stable. To catch issues in clang dev version, add clang-18 variant for llvm build on x86-64.


chantra:
let's wait for the tests to confirm this is not blowing up, but otherwise LGTM.

chantra (approved):

PR #255, opened 2023-12-11
Use dataclass to define build configs

yurinnick:
Instead of relying on dict editing dark magic, define `BuildConfig` class that includes all fields required for build customization and then generate build matrix out of it.


chantra (approved):
LGTM. Let's wait for testing to finish though.

PR #261, opened 2024-02-12
build: Always pass architecture to actions

chantra:
In preparation of cross-compiling, ensure that we always pass arch parameter to the action.
At a later stage, this parameter will be made required on libbpf-ci.


chantra:
https://github.com/kernel-patches/vmtest/actions/runs/7879475334/job/21499804418?pr=261#step:11:4

shows that we get a warning for the new parameter, but other than that, libbpf/ci happily ignore the parameter.

PR #272, opened 2024-04-23
ci/diffs: Add temporary fix for mitigation config

liu-song-6:
Upstream is discussing the exact config to ship. In the meanwhile, which would unblock CI.

More discussions here:

https://lore.kernel.org/lkml/20240423045548.1324969-1-song@kernel.org/T/#u


chantra:
LGTM

PR #274, opened 2024-05-29
ci: build: Make -O2 build name a 3-tuple as well

danobi:
All other job names are in 'X for Y with Z' form. Normalize the build job name to that form as well for easier parsing.


chantra (approved):
LGTM

PR #277, opened 2024-08-06
Add DENYLIST for RC tier

danobi:
Production CI runs on bare-metal self hosted runners. RC CI runs on nested virt GH hosted runners. The difference in platform causes issues with some selftests. So add a DENYLIST to skip tests that are too difficult to debug.

Tested here: https://github.com/kernel-patches/bpf-rc/pull/4201


danobi:
cc @yonghong-song

PR #279, opened 2024-08-12
ci: skip test_maps on s390x

anakryiko:
It's flaky and OOMs pretty frequently. We get lots of signal on x86-64, so let's disable it on s390x.


danobi (approved):
would be nice to leave inline comment about why as well

anakryiko:
> would be nice to leave inline comment about why as well

sorry, pushed before seeing the comment. Well, at least git blame will point to commit message, so probably not a big deal.

PR #286, opened 2024-09-13
workflow: Also run the workflow against bpf-net_base and for-next_base

chantra:
We have added those 2 branches to be tracked. For CI to run when those get updated, we need to add them to the list of branches to track.


theihor (approved):
LGTM

chantra:
We are hitting rate limiting issues, so I will probably hold a bit on this one.

PR #309, opened 2024-11-26
kernel-build: always try incremental kernel build

theihor:
Remove CACHED_KERNEL_BUILD variable. Instead, always try cache lookup and an incremental kernel build.

Fix base-branch passed to prepare-incremental-build action step:
  * use BPF_NEXT_BASE_BRANCH (master) if linux source is downloaded
  * use base_ref for pull_requests
  * default to ref_name otherwise

Clear build artifacts to avoid caching them on non-push events.


theihor:
Also fixes rc failures: 
* https://github.com/kernel-patches/bpf-rc/actions/runs/12038145531
* https://github.com/kernel-patches/bpf-rc/actions/runs/12038144471

PR #320, opened 2024-12-10
Re-enable sched_ext test

theihor:
Include an upstream patch for selftests/sched_ext

Link: https://lore.kernel.org/all/20241209152924.4508-1-void@manifault.com/


theihor:
Added https://lore.kernel.org/bpf/Z1n9v7Z6iNJ-wKmq@slm.duckdns.org/

PR #337, opened 2025-02-03
GCC BPF: use pre-built release

theihor:
See the commit message.


theihor:
The test was successful, but I don't see a good reason to use this approach for now.

theihor:
Ok, actually we might want to use this in BPF CI.

The problem with current implementation is that many copies of the same GCC build are cached, because caches between pull requests are separate. So using a pre-built release is much more reasonable.

I am going to restore and update this change to use latest release at https://github.com/theihor/gcc-bpf

PR #348, opened 2025-03-04
matrix.py: configure codebuild runners

theihor:
Ondemand AWS CodeBuild runners only make sense for kernel-patches/bpf, as this is the main CI repository. In particular, there is no reason to use CodeBuild on bpf-rc.

Control this in the matrix.py script: set build_runs_on to "codebuild" for managed repositories.


theihor:
Tested on bpf-rc: https://github.com/kernel-patches/bpf-rc/actions/runs/13665318206

PR #360, opened 2025-04-03
Revert "ci/diffs: add fixes from bpf tree"

theihor:
This reverts commit 0925e82ead4b41ee89ba26f9e8c4118d13d49d6c.


theihor:
GCC job fails with obscure error when downloading Linux source snapshot. Ignoring that.

PR #369, opened 2025-04-30
ci: refactor toolchain setup

theihor:
The 'toolchain' matrix parameter has been used in a confusing way. The
'version' field was only applicable to LLVM, but there is also a
separate 'llvm-version' action parameter spread across the workflows.

Now that we want to be able to specify a particular GCC version for
kernel build, a refactoring is appropriate.

Given that libbpf/ci/setup-build-env action sets up a particular
compiler version systemwide, and mixing LLVM versions between kernel
and selftests build is an unlikely scenario, it makes sense to simply
have gcc_version and llvm_version parameter, fixed within a particular
workflow. And so the composite 'toolchain' becomes unnecessary: it's
enough to distinguish between gcc and llvm build of the kernel.

However 'toolchain' and 'toolchain_full' are still useful variables
between the workflows and actions, particularly for identifying build
artifacts.

Here is what happens now:
  * DEFAULT_{GCC,LLVM}_VERSION is set in matrix.py
  * 'toolchain' is replaced with 'kernel_compiler' in the job matrix
  * test.yml workflow looks at kernel_compiler to construct and pass
   'toolchain' and 'toolchain_full' to other workflows
  * '{gcc,llvm}_version' inputs are used to determine compiler
    versions for the build setup


theihor:
Note: this depends on https://github.com/libbpf/ci/actions/runs/14762852701

PR #386, opened 2025-06-30
ci/diffs: clean up unnecessary patches

theihor:
I used github cli tool to fetch the logs on the base branches to determine which ci/diffs are actually applied.

Only three patches are still used at this time:
  * removal of llvm check in bpftool
  * a selftests fix that probably should either be removed or landed upstream
  * recent cgroup xattr fix, still needed on for-next branch???


theihor:
For reference, a piece of bash to figure out which ci/diffs are applied:
```bash
for BRANCH in "bpf_base" "bpf-next_base" "for-next_base" "bpf-net_base"; do
    RUN_ID=$(gh run list --repo "$REPO" --branch "$BRANCH" --status success --limit 1 --json databaseId -q '.[].databaseId')
    gh run view "$RUN_ID" --repo "$REPO" --json jobs > $json
    cat $json | jq '.jobs[] | select(.steps[].name | contains("patch-kernel")) | .databaseId' > $job_ids
    for JOB_ID in $(cat $job_ids); do
        log="${OUT}/${RUN_ID}-${JOB_ID}.log"
        echo "Fetching $log"
        gh api repos/$REPO/actions/jobs/$JOB_ID/logs > $log
        # gh run view $RUN_ID --job=$JOB_ID --repo $REPO --log > $log
    done
done

grep 'Successfully applied.*' -roh | grep -o 'ci/diffs/.*.patch' | sort -u
```

I don't think it's worth committing anywhere.

PR #419, opened 2025-12-04
ai-code-review.yml: switch to Claude Opus 4.5

theihor:


theihor:
Something is wrong when switching to new model id: https://github.com/kernel-patches/bpf-rc/actions/runs/19946377031

Needs investigation...

theihor:
The workflows didn't have access due to [changes on Amazon side](https://aws.amazon.com/blogs/security/simplified-amazon-bedrock-model-access/). Fixed in AWS account.

PR #441, opened 2026-02-05
Implement a workflow to run veristat on cilium programs

puranjaymohan:


.github/workflows/veristat-cilium.yml
```diff
@@ -0,0 +1,74 @@
+name: veristat_kernel
```
theihor: "veristat cillium"?

theihor (changes requested):

theihor (approved):

PR #502, opened 2026-07-31
Gcc 16 switch

vineetgarc:


theihor:
Hi @vineetgarc . The error highlighted in the github UI is red herring. If you download the raw logs you'll get 469M log with this at the end:
```
 0x7ff)) R8=scalar(id=79999,smin=umin=smin32=umin32=0x13878,smax=umax=smax32=umax32=0x13c5f,var_off=(0x13800; 0x7ff))
2026-08-01T19:03:11.9899382Z 243: (bf) r2 = r8                     ; R2=scalar(id=79999,smin=umin=smin32=umin32=0x13878,smax=umax=smax32=umax32=0x13c5f,var_off=(0x13800; 0x7ff)) R8=scalar(id=79999,smin=umin=smin32=umin32=0x13878,smax=umax=smax32=umax32=0x13c5f,var_off=(0x13800; 0x7ff))
2026-08-01T19:03:11.9899477Z 244: (bf) r1 = r6                     ; R1=arena R6=arena
2026-08-01T19:03:11.9899555Z 245: (85) call pc+125
2026-08-01T19:03:11.9899657Z Func#3 ('htab_update_elem') is global and assumed valid.
2026-08-01T19:03:11.9899721Z 246:
2026-08-01T19:03:11.9899784Z ;  @ arena_htab.c:46
2026-08-01T19:03:11.9899853Z 246: (bf) r7 = r8
2026-08-01T19:03:11.9899954Z BPF program is too large. Processed 1000001 insn
2026-08-01T19:03:11.9900178Z processed 1000001 insns (limit 1000000) max_states_per_insn 65 total_states 80008 peak_states 80008 mark_read 0
2026-08-01T19:03:11.9900281Z -- END PROG LOAD LOG --
2026-08-01T19:03:11.9900391Z libbpf: prog 'arena_htab_asm': failed to load: -E2BIG
2026-08-01T19:03:11.9900490Z libbpf: failed to load object 'arena_htab_asm'
2026-08-01T19:03:11.9900634Z libbpf: failed to load BPF skeleton 'arena_htab_asm': -E2BIG
2026-08-01T19:03:11.9900795Z test_arena_htab_asm:FAIL:arena_htab_asm__open_and_load unexpected error: -7
2026-08-01T19:03:11.9900885Z #3/2     arena_htab/arena_htab_asm:FAIL
```

Please make sure the CI here is green. I suspect you need to adjust the denylist more. It's ok for it to be big at this point.

theihor:
btw, if you could come up with an (independent) fix for the misleading error that would be great too.

The problem is that the script running the test_progs in a VM was killed due to the timeout, and so output post-processing didn't run, which is why we get this error downstream. The code doing all this is here: https://github.com/libbpf/ci/blob/main/run-vmtest/run.sh

.github/scripts/matrix.py
```diff
@@ -188,10 +188,8 @@ def tests(self) -> Dict[str, Any]:
         # if self.arch in [Arch.X86_64, Arch.AARCH64] and not self.is_netdev:
         #    tests_list.append("sched_ext")
 
-        # Don't run GCC BPF runner, because too many tests are failing
-        # See: https://lore.kernel.org/bpf/87bjw6qpje.fsf@oracle.com/
-        # if self.arch == Arch.X86_64:
-        #    tests_list.append("test_progs-bpf_gcc")
+        if self.arch == Arch.X86_64:
```
theihor: We need to preserve the "if netdev" check here. We don't need to run gcc-bpf for netdev. The job that you deleted enforced `!inputs.is_netdev`. With this change every x86 netdev configuration now builds and schedules `test_progs-bpf_gcc`.

.github/workflows/kernel-build.yml
```diff
@@ -120,11 +120,25 @@ jobs:
...
+          GCC_BPF_VERSION: ${{ vars.GCC_BPF_VERSION || '16' }}
+        run: .github/scripts/download-gh-release.sh theihor/gcc-bpf ${{ github.workspace }}/gcc-bpf "gcc-${GCC_BPF_VERSION}-"
+
       - name: Build selftests/bpf
         uses: libbpf/ci/build-selftests@v4
         env:
           MAX_MAKE_JOBS: 32
           RELEASE: ${{ inputs.release && '1' || '' }}
+          # Build test_progs-bpf_gcc on x86_64 so the vmtest runner has a binary to run.
+          BPF_GCC: ${{ inputs.arch == 'x86_64' && format('{0}/gcc-bpf', github.workspace) || '' }}
```
theihor: There is a problem with moving gcc-bpf path within the main kernel-build workflow: if something breaks with the scripts or selftests build, we now failing the "root" build job that every other test run depends on. Separate gcc-bpf workflow avoided that.

I think we can still delete gcc-bpf.yml, but then gcc-specific steps need to be allowed to fail and produce a separate artifact or something like that.

Think about how to do this. The main point is to keep gcc-bpf isolated: if it's broken, don't fail other independent jobs.

.github/workflows/kernel-build.yml
```diff
@@ -120,11 +120,25 @@ jobs:
...
+      - if: ${{ inputs.arch == 'x86_64' }}
+        name: Download GCC BPF compiler
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
+          # Major GCC version used to compile the BPF selftest programs. The
+          # latest theihor/gcc-bpf release of that series is used. Override with
+          # the repo variable GCC_BPF_VERSION (Settings -> Secrets and variables
+          # -> Actions -> Variables) to try another series without a code change.
+          GCC_BPF_VERSION: ${{ vars.GCC_BPF_VERSION || '16' }}
```
theihor: I'd avoid having a default value here. TBH I don't think we need to track different releases for gcc-bpf at all: we only care about testing the "latest" one, so why bother? It's just an additional surface for bugs.

.github/scripts/download-gh-release.sh
```diff
@@ -6,16 +6,27 @@ SCRIPT_DIR=$(dirname "$(realpath "$0")")
...
 
 cd /tmp
 
 bash "$SCRIPT_DIR/install-github-cli.sh"
 
-tag=$(gh release list -L 1 -R ${GH_REPO} --json tagName -q .[].tagName)
+if [[ -n "$TAG_PREFIX" ]]; then
+    # Releases are listed newest first, so the first match is the latest one.
+    tag=$(gh release list -L 100 -R "${GH_REPO}" --json tagName -q '.[].tagName' \
+              | grep -m1 -- "^${TAG_PREFIX}" || true)
```
theihor: looks like TAG_PREFIX is read as a regex

I think it should either be called "_REGEX" or be matched strictly as prefix

theihor (changes requested):

vineetgarc:
> btw, if you could come up with an (independent) fix for the misleading error that would be great too.

OK that's addressed via [1] and also a fix [2] in selftests to not generate those names in the first place.

[1] https://github.com/kernel-patches/vmtest/pull/504
[2] https://lore.kernel.org/bpf/fcfe4ed2-5c04-45df-a4b0-27d1bfddf986@linux.dev/T/#m06523f76ae187fbf6962777cb6961e97d715d55b

.github/workflows/kernel-build.yml
vineetgarc: I don't have a strong preference for deleting it - Claude suggested it was redundant after the changes. we can keep it and rework the patches along those lines.

.github/workflows/kernel-build.yml
vineetgarc: I think it is still useful, even if only occasionally needed, to build something with a specific GCC release driven by a UI toggle.

.github/workflows/kernel-build.yml
vineetgarc: Ok so there are two plausible approaches:

Option 1 — self-contained gcc-bpf job (build and run in one job)
  gcc-bpf.yml already downloads the vmlinux artifact, fetches the compiler, and builds test_progs-bpf_gcc. It just
  throws the binary away. This option adds a run-vmtest step so the same job runs it.

Option 2 — gcc-bpf.yml builds and uploads; the test job downloads it
  Keeps today's job split, but the binary is published instead of discarded.

Second option causes a delay of every job behind gcc-bpf it so not preferable.
But option 1 then wants a KVM runner (vs. builder runner) for gcc-bpf which is not the case currently.
Its a trade-off as always, which one do we prefer.

vineetgarc:
> Hi @vineetgarc . The error highlighted in the github UI is red herring. If you download the raw logs you'll get 469M log with this at the end:
> 
...
> 2026-08-01T19:03:11.9900885Z #3/2     arena_htab/arena_htab_asm:FAIL
> ```
> 
> Please make sure the CI here is green. I suspect you need to adjust the denylist more. It's ok for it to be big at this point.

Yep added this test to denylist !

vineetgarc:
Superseded by https://github.com/kernel-patches/vmtest/pull/505

PR #504, opened 2026-08-06
kernel-test: normalize traffic monitor log names before uploading

vineetgarc:
The traffic monitor names its logs after the test and subtest, so they can contain characters that actions/upload-artifact rejects. A subtest called "INET4: bpf_timestamping" makes the upload fail as below (see [1])

  Error: The path for one of the files in artifact is not valid:
  /packets-125-15-net_timestamping__INET4:_bpf_timestamping-net_timestamping_ns.log.
  Contains the following character:  Colon :

run-vmtest already renames these, but only after the VM exits, so a step timeout or a dead VM kills it before that point. The upload step runs with if: always() and then trips over the names that were left behind.

Rename them in a step of its own, also guarded by if: always(), so it happens whatever the outcome of the test step. It is a no-op when the directory is absent, which is the case for every test that does not enable the traffic monitor, and re-running it is harmless.

Link: https://github.com/kernel-patches/vmtest/actions/runs/30710914503/job/9139901138 [1]

### DIFF
--- a/.github/workflows/kernel-test.yml
+++ b/.github/workflows/kernel-test.yml
@@ -94,6 +94,22 @@ jobs:
           # it was extracted from the artifacts.
           kbuild-output: ${{ env.REPO_ROOT }}/kbuild-output
 
+      # run-vmtest normalizes these names itself, but it does so after the VM
+      # exits, so a timeout or a dead VM kills the step before it gets there.
+      # The upload below runs regardless, so redo it here to be safe.
+      - if: ${{ always() }}
+        name: Normalize traffic monitor log names
+        shell: bash
+        run: |
+          dir=/tmp/tmon_pcap
+          [ -d "$dir" ] || exit 0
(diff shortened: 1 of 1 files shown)


theihor (changes requested):
This essentially copy-pastes [a script](https://github.com/libbpf/ci/blob/main/run-vmtest/normalize-paths-for-github.sh) from libbpf/ci/run-vmtest, which is an unnecessary duplication.

`always()` is the right thing to do, but I think the right spot for it is in [run-vmtest/action.yml](https://github.com/libbpf/ci/blob/main/run-vmtest/action.yml) as a last step.

So basically we need factor out this normalization from [run.sh](https://github.com/libbpf/ci/blob/main/run-vmtest/run.sh#L118-L120) into action.yml.

vineetgarc:
This is superseded by https://github.com/libbpf/ci/pull/231

PR #505, opened 2026-08-07
Gcc 16 switch and re-enable

vineetgarc:
This re-enables GCC built selftests - uses latest GCC (16 as of now)
 - denylist refreshed for current gcc-16 based builds
 - gcc-bpf: run the selftests instead of just building them (runner was disabled back in 2025)
   + This is more involved as it requires splitting into build and run jobs (see that patch for more details)


theihor:
@vineetgarc are you still working on this? The gcc-bpf job is red, the PR doesn't look ready for review.

.github/workflows/gcc-bpf.yml
```diff
@@ -26,16 +26,22 @@ on:
...
       ARCH: ${{ inputs.arch }}
+      ARTIFACTS_ARCHIVE: "selftests-bpf-gcc-${{ inputs.arch }}-${{ inputs.toolchain_full }}.tar.zst"
       BPF_NEXT_BASE_BRANCH: 'master'
       GCC_BPF_INSTALL_DIR: ${{ github.workspace }}/gcc-bpf
       GCC_BPF_RELEASE_REPO: 'theihor/gcc-bpf'
+      # Major GCC version used to compile the BPF selftest programs. The latest
+      # release of that series is used. Override with the repo variable
+      # GCC_BPF_VERSION (Settings -> Secrets and variables -> Actions ->
+      # Variables) to try another series without a code change.
+      GCC_BPF_VERSION: ${{ vars.GCC_BPF_VERSION || '16' }}
```
theihor: Ok, if we want to keep the version toggle, let's do it more explicitly. Currently only the default is used and the var is never actually set: vars is github repo or org level - not a good place for compiler version value.

Let's make it:
```
GCC_BPF_RELEASE_PREFIX: 'gcc-16-'
```
which is then passed directly to `download-gh-release.sh`

ci/vmtest/configs/DENYLIST.test_progs-bpf_gcc
```diff
@@ -1,45 +1,38 @@
-arena_htab
```
theihor, line 1: If we are keeping the gcc-bpf version toggle, should we support multiple denylists? e.g. `DENYLIST.test_progs-bpf_gcc-{15,16}`

.github/workflows/kernel-test.yml
```diff
@@ -94,6 +94,22 @@ jobs:
...
+        shell: bash
+        run: |
+          dir=/tmp/tmon_pcap
+          [ -d "$dir" ] || exit 0
+          find "$dir" -depth -name '*[":<>|*?]*' -print0 |
+          while IFS= read -r -d '' path; do
+            base=$(basename "$path")
+            new="$(dirname "$path")/$(printf '%s' "$base" | tr '":<>|*?\r\n' '_________')"
+            [ "$path" = "$new" ] || mv -v -- "$path" "$new"
+          done
```
theihor: As I [commented already,](https://github.com/kernel-patches/vmtest/pull/504#pullrequestreview-4910890691) let's fix this on libbpf/ci/run-vmtest side.

.github/workflows/kernel-build-test.yml
theihor, line 187: We are currently running gcc-bpf workflow both for the kernel built with llvm and with gcc. I'm thinking that may be an overkill.

How about we add `&& inputs.toolchain == 'gcc'` here?

ci/vmtest/configs/DENYLIST.test_progs-bpf_gcc
```diff
@@ -1,45 +1,38 @@
...
-async_stack_depth
+arena_htab/arena_htab_asm           # 1M runaway since v7.2-rc1
+arena_strsearch
 bad_struct_ops/invalid_prog_reuse
```
theihor, line 3: nit: In the commit message you say:

> The denylist shrinks from 904 to 193 entries (61 whole tests, 132 subtests). A

Let's not put the exact numbers there, they will almost certainly drift between revisions.

.github/workflows/gcc-bpf.yml
```diff
@@ -26,16 +26,22 @@ on:
...
-  test:
-    name: GCC BPF
+  build:
+    name: GCC BPF build
     runs-on:
       - ${{ format('codebuild-bpf-ci-{0}-{1}', github.run_id, github.run_attempt) }}
       - image:custom-linux-ghcr.io/kernel-patches/runner:kbuilder-debian-x86_64
     env:
       ARCH: ${{ inputs.arch }}
+      ARTIFACTS_ARCHIVE: "selftests-bpf-gcc-${{ inputs.arch }}-${{ inputs.toolchain_full }}.tar.zst"
```
theihor: This should be an absolute path. Use `${{ github.workspace }}`

.github/workflows/gcc-bpf.yml
```diff
@@ -85,7 +91,7 @@ jobs:
         shell: bash
         env:
           GH_TOKEN: ${{ github.token }}
-        run: .github/scripts/download-gh-release.sh ${{ env.GCC_BPF_RELEASE_REPO }} ${{ env.GCC_BPF_INSTALL_DIR }}
+        run: .github/scripts/download-gh-release.sh ${{ env.GCC_BPF_RELEASE_REPO }} ${{ env.GCC_BPF_INSTALL_DIR }} "gcc-${GCC_BPF_VERSION}-"
```
theihor: `${}` and `${{}}` are mixed up in one line. Use `${{ env.VAR }}` here.

.github/workflows/gcc-bpf.yml
```diff
@@ -98,3 +104,78 @@ jobs:
...
+          VMTEST_CONFIGS: ${{ github.workspace }}/ci/vmtest/configs
+          VMTEST_MEMORY: 5G
+          VMTEST_NUM_CPUS: 4
+          TEST_PROGS_WATCHDOG_TIMEOUT: 600
+        with:
+          arch: ${{ inputs.arch }}
+          vmlinuz: '${{ github.workspace }}/vmlinuz'
+          kernel-root: ${{ env.REPO_ROOT }}
+          max-cpu: 8
+          kernel-test: test_progs-bpf_gcc
```
theihor: Let's add `KERNEL_TEST: test_progs-bpf_gcc` to the top and use it in both places.

theihor (changes requested):
For the next respin, please use nice PR title and add a brief summary too. I now see that it depends on https://github.com/libbpf/ci/pull/228, would be nice to mention that.

To submit a new revision you can just force push updated `gcc-16-switch-n-fix-pathname` branch, PR will be refreshed. Don't close and re-open the PR, it's harder to track the comments then.

Could you please also delete a commented block in [matrix.py](https://github.com/kernel-patches/vmtest/blob/master/.github/scripts/matrix.py#L191-L194)? With this PR it's stale (uncommenting won't work).

vineetgarc:
> @vineetgarc are you still working on this? The gcc-bpf job is red, the PR doesn't look ready for review.

Yes I'm. This was gated on the libbpf-ci 228 which you pulled. Let me see what other changes you requeted and respin with better PR look and feel. Thx.

ci/vmtest/configs/DENYLIST.test_progs-bpf_gcc
vineetgarc: Umm good point. So lets just drop support fo rold version as that denylist itself might be amoving target given kernel changes and will technically need updating.

vineetgarc: This also implies that theihor/bpf-gcc now need to publish only gcc-16 ? So I'll first have to make that merge. Agree ? It should not matter because today there are no consumers of whatever it generates.

.github/workflows/gcc-bpf.yml
vineetgarc: Fixed !

.github/workflows/gcc-bpf.yml
vineetgarc: I think this will go away with hardcoded gcc-16

.github/workflows/kernel-test.yml
vineetgarc: Done. https://github.com/libbpf/ci/pull/231 !

ci/vmtest/configs/DENYLIST.test_progs-bpf_gcc
vineetgarc: OK !

ci/vmtest/configs/DENYLIST.test_progs-bpf_gcc
theihor: There is a consumer: [current gcc-bpf workflow](https://github.com/kernel-patches/bpf/actions/runs/31541334338/job/93947642036).

I think your changes to theihor/gcc-bpf and the download-gh-release.sh script are ok. The script pulls latest by default. All we need to do is configure theihor/gcc-bpf to only build gcc 16 and that's it. As soon as this is done, the current BPF CI will pick up new compiler, but that should be ok - it only builds the tests anyways.

theihor: oh you mean consumer for the objects... yes, you're right about that

.github/workflows/kernel-build-test.yml
vineetgarc: IMHO this is a valid coverage point !

vineetgarc:
@theihor fwiw this series has been proven to pass: cfr https://github.com/kernel-patches/vmtest/actions/runs/31639638784 
We are missing  the libbpf-ci version tag v4 to be moved to encompass 386733ce0455 normalize_bpf_test_names: don't drop tests that share a name prefix (#228

theihor:
@vineetgarc A couple of things to address:

- [bpf_cookie/perf_event is failing](https://github.com/kernel-patches/vmtest/actions/runs/31651855019/job/94300939728), maybe flaky
- we have a `run_tests` flag [passed through](https://github.com/kernel-patches/vmtest/blob/25eafb4dcd3089d6705f099ea3fc14099ba316fd/.github/workflows/kernel-build-test.yml#L99) from matrix.py, I think you should use it for the test step as well
- and looks like you haven't addressed [my first comment](https://github.com/kernel-patches/vmtest/pull/505#pullrequestreview-4911064324): please fixup matrix.py and add brief summary and proper title for the PR

For testing, add a commit on top that renames `libbpf/ci/run-vmtest@v4` to `libbpf/ci/run-vmtest@main` in `gcc-bpf.yml`. Then you can just drop it when the PR is ready for merge.

I don't want to move the tag right now, because CI is busy due to recent pushes to bpf and bpf-next. I expect tomorrow to be quieter.

Thanks.

vineetgarc:
> @vineetgarc A couple of things to address:
> 
> * [bpf_cookie/perf_event is failing](https://github.com/kernel-patches/vmtest/actions/runs/31651855019/job/94300939728), maybe flaky

Indeed it is flaky in my local runs: added now.

> * we have a `run_tests` flag [passed through](https://github.com/kernel-patches/vmtest/blob/25eafb4dcd3089d6705f099ea3fc14099ba316fd/.github/workflows/kernel-build-test.yml#L99) from matrix.py, I think you should use it for the test step as well

Done.

> * and looks like you haven't addressed [my first comment](https://github.com/kernel-patches/vmtest/pull/505#pullrequestreview-4911064324): please fixup matrix.py and add brief summary and proper title for the PR

Oops done now. !

> For testing, add a commit on top that renames `libbpf/ci/run-vmtest@v4` to `libbpf/ci/run-vmtest@main` in `gcc-bpf.yml`. Then you can just drop it when the PR is ready for merge.

I think the prereq it wants is there. The 2nd one is not really mandatory. I'll check anyways.

> I don't want to move the tag right now, because CI is busy due to recent pushes to bpf and bpf-next. I expect tomorrow to be quieter.

No rush.

Thx !

theihor:
It's green, congrats! :)

I noticed a bunch of errors like this in the logs:
```
2026-08-14T16:20:24.4439743Z libbpf: Error loading .BTF into kernel: -22. BTF is optional, ignoring.
2026-08-14T16:20:32.0898416Z [  136.250755] ip6_tunnel: gre6_dev xmit: Local address not yet configured!
2026-08-14T16:20:32.5270475Z libbpf: BTF loading error: -22
```

Is this a known issue? The test_progs reports PASS on everything, so maybe it's not a blocker for the CI changes.

theihor (approved):

vineetgarc:
Woo hoo ! and in this process uncovered a legit x86 jit bug ! https://lore.kernel.org/bpf/20260814165557.3405518-1-vineet.gupta@linux.dev/T/#u

PR #525, opened 2026-09-10
ci: add a new workflow to test bpftool build

Tropicao:
The tools/testing/selftests/bpf directory in the kernel source tree contains the test_bpftool_build.sh script that ensures that the various supported ways of building bpftool work correctly. This test isn't currently part of the tests automatically executed in CI.

Add a new layer of workflows in test.yml to run tests that will exercise build commands. As a first sub-workflow, add a bpftool build workflow:
  test.yml -> test-build-commands.yml -> build->bpftool.yml
This test-build-commands workflow can be extended later to support other artifacts build test, for example to test the different ways of building selftests.

The new build-bpftool workflow aims to validate the different ways of building the bpftool CLI, as exercised by the test_bpftool_build.sh in the kernel source tree (in tools/testing/selftests/bpf). As the test has only a few dependencies (only a build test, no dependency on built artifact), the added workflow remains simple and isolated from the other "build-and-tests" workflows:
- it only tests host build, so no need to run it as many time as the matrix script generates configurations
- supports download_sources flag, so can be run from kernel tree or vmtest tree

---
This new workflow does not make the whole test_bpftool_build.sh execute in CI; as some subtests are gated by the presence of a .config. I suspect this to be hiding a broader kbuild issue as bpftool does not really depend on .config, I'll fix this separately in another series.

### DIFF
--- a/.github/workflows/build-bpftool.yml
+++ b/.github/workflows/build-bpftool.yml
@@ -1,0 +1,46 @@
+name: Test bpftool build methods
+
+on:
+  workflow_call:
+    inputs:
+      download_sources:
+        required: true
+        type: boolean
+        description: Whether to download the linux sources into the working directory.
+
+jobs:
+  build-bpftool:
+    name: Test bpftool build
+    timeout-minutes: 10
+    permissions: read-all
+    runs-on: ubuntu-latest
+    env:
+      BPF_NEXT_BASE_BRANCH: 'master'
+      REPO_ROOT: ${{ github.workspace }}/src
+    steps:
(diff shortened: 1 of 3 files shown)


theihor:
@Tropicao the CI run with this change [hasn't even started](https://github.com/kernel-patches/vmtest/actions/runs/34488684980).

The problem is probably with `permissions: read-all`: the workflow that is called inherits permissions of the caller, and can not expand them (and in this case read-all does).

theihor (changes requested):
Please respin.

